PR #3, opened 2022-07-12
expose created/modified timestamps as columns in magic.mse-game

haganbmj:
Unsure what's the more legible way to name those columns, so it could be better to stick with the internal names `Time Created` and `Time Modified`.


CajunAvenger:
these columns can already be added via the column menu? that's why they're in card fields in the first place

haganbmj:
Ah whoops nevermind - I think I was confusing `card list visible` and `card list allow`. It's `allow` that controls whether they're able to added and that defaults to `true`.

https://github.com/twanvl/MagicSetEditor2/blob/5e50cb4c56cd07b9aa97d82efbe0d2dce5c9ab5c/src/data/field.cpp#L33-L34

PR #9, opened 2023-01-24
Cockatrice export v2

ebbit1q:
update to xml v4
escape special xml characters in cockatrice exporter
indent xml
update icon
rename to 2.0


ebbit1q:
I've tested these to the best of my abilities but it's likely I have not come across all of the edge cases, I suggest testing some more elaborate exports before merging

CajunAvenger:
Have done the main pass:

- Fixed a problem in maintype() that could break cards with an unexpected type

- Added support for `persistent` tag
-- Found out that doesn't work properly in Cockatrice on reverse-related, investigating i think CardDatabase::refreshCachedReverseRelatedCards() is clearing the persistent flag, will need to hit that when I do the PR for updating Oracle now that MTGJson has spellbooks.
- added !token to support <related> tags, which can be a temp backup for persistent cards as well
- Removed requirement for !related to end in a semicolon that 1.04 introduced
- Condensed options so they're all visible on the starting screen

The way we add token relations does break your indenting so I'm gonna try to clean that up tomorrow, only other thing would be considering renaming this, not sure CS/ECH is even active anymore.

ebbit1q:
good catch

would that need to be addressed in cockatrice? we can look into that, would you like to ticket it?

cool

makes sense

looks good

I might have missed that, didn't test it in enough depth I guess. I didn't know what CS/ECH was so I didn't change that, I'd say it should just be called cockatrice exporter v2.

CajunAvenger:
> would that need to be addressed in cockatrice? we can look into that, would you like to ticket it?

it does yea, looking over the files refreshCachedReverseRelatedCards seems to be the trouble point with redoing the card relations with the default (off) persistent tag, i had just missed that it doing that on my original PR.

it's not a critical issue as it still works fine with normal related, so probably i'll just bundle it with the Oracle support for canon conjure cards i was gonna do anyway.

ebbit1q:
the exporter right now does not do normal card relations though, for some reason it only does reverse-related and the !related command is named quite confusingly, I'm not sure what the best approach to fix that would be, I don't want to break people's existing !related commands in their sets.

ebbit1q:
ah, I see in the diff of your commit you've added the !token command, that fixes that, though it's still named confusingly.

CajunAvenger:
yea related vs reverse-related we're a few years late on at this point and probably just stuck with

!tokens is named cause the usual use-case is that you do `!tokens This;Card's;Tokens`, which is a feature request that's been thrown at the exporter a few times. Persistent cards piggybacking on it is mostly just a temporary measure

ebbit1q:
also, before reworking oracle consider downloading xmls directly, not relying on oracle's hacky way of parsing which makes it eat up all your machine's ram: https://github.com/Cockatrice/Cockatrice/pull/4736
I'm hoping to set up some infrastructure to stop wasting people's ram someday.

PR #12, opened 2023-03-06
Statistics Update

G-e-n-e-v-e-n-s-i-S:
Made all card fields "show statistics: false"
Changed a few card field icons
Added "card stamp 2" field
Added statistics section to set fields
Overhauled statistics page
Changed "has_identity" method in the vanguard template


CajunAvenger:
mkay pretty sure there's a way to clean out the unused stats i'm gonna look at that to make things less cramped

all the extra color density stuff i'm probably gonna move further down the list

CajunAvenger:
for the color pips stats, we should probably do one of two things:

a) use the zero-width space trick to alphabetize the colors, letting us not post all the empty slots

b) instead have a statistic dimension that's just color pip count, such that you can apply Cast. Costs Pips Total and Cast. Costs Pip Density to get a stacked bar graph of W|WW|WWW|WWWW. multicolor might make that more annoying than the zero-width shenanigans


for the sub type statistics, i'm going to change how word-lists is built so that this and word-lists can build from the same source instead of needing to keep track of two instances. annoyingly we can't reference word-lists directly, but we can build them via a script and just have them both reference the same script

G-e-n-e-v-e-n-s-i-S:
For the color pips stats, I'll implement solution a)
Solution b) seems like it would be missed by some users, they would not realize you could do that.

For the type lists, I was about to suggest that. However if we want to specify colors, we still need to have a copy of the lists in the statistics file. But removing the extra copy in the statistics_script file is worth it.

CajunAvenger:
> I currently replace these spaces with EN spaces

o so that's how you fixed mythic rare. good to know

> Some types are a part of other types, like Serra and Serra's Realm. So, need to avoid false positives with stuff like contains(match:"Serra"). This is why I add a semicolon at the end of each type in the statistics_scripts lists. I can search for "Serra;" instead of "Serra".

yep, word_list needs those to be commas so i'll either switch those over or have the space-changer also change commas to semicolons

CajunAvenger:
couple notes from the last run through

- card shape *should* always be lowercase, could simplify your double checks there
- `face_detect_custom_creature_subtypes_statistic` seems to have been way overdesigned? i simplified that and i'm pretty sure all those supertype removal closures can be removed now
- i saw a few comments about having trouble with card_style initialization, have you tried the tech of `if card_style[whatever] or else -1 > 0` to dodge errors on init, or is it just unable to access what you're after at all?

beyond that i thiiink i'm done with the main code review, though i am going to start stress testing Filter to make sure it doesn't crash or lock to user error too easily

G-e-n-e-v-e-n-s-i-S:
- ```face_detect_custom_creature_subtypes_statistic``` looks for cards that are creatures with no other type. that way it can infer that all subtypes are creature subtypes. if a card is an artifact creature, it can't tell if the custom subtypes are creature subtypes or artifact subtypes. however, this whole process inevitably misses a few, and can't tell if it's a race subtype or a class subtype, so maybe it's too janky to justify it's existence in the first place?
 - i tried the ```if card_style[whatever] or else -1 > 0``` trick but couldn't get it to work. I don't understand the behavior of ```card_style``` while looping over the cards in a stats script. it seems to use the value of the last initialized card? I admit I gave up, maybe you can give it a go.
 - for the renaming of the stats, I think the word "Density" is not descriptive enough. this morning I looked at the ```casting cost density``` stat and for a minute I had no idea what it was counting... and I wrote the thing. Using "Casting Cost Pip Count" is more explicit, even if it is a tiny bit too long to fit in the column.
 - for the creature word lists: we could generate the "frequently used" list at the start through a script as well. this could then double as being the entire list for languages that do not use the latin alphabet.
 - for localization in general, I have added a few things to the language file, translations of common types, number maps, and a few function definitions that I used when I translated the keywords. this probably needs to be in another branch but I'm just gonna leave the file here if you want to take a look.
[language.txt](https://github.com/MagicSetEditorPacks/Full-Magic-Pack/files/13278369/language.txt)

CajunAvenger:
> looks for cards that are creatures with no other type. that way it can infer that all subtypes are creature subtypes.

ic. i'll have to tank on how we might make that more reliable

> for the renaming of the stats, I think the word "Density" is not descriptive enough

pfft i had the exact opposite where i thought it was some overlooked duplicate from just the name.

> for the creature word lists: we could generate the "frequently used" list at the start through a script as well

technically it does this already, the top word of the main lists are a script that collects all of them you've used in this set file. For non-Latin languages, i'll be setting up alternate word lists, so they'll use like `race-ja` instead of `race` which can then get sorted more naturally for that language

> for localization in general, I have added a few things to the language file

ooo now you're cooking. this should be another branch yea, one i'll tackle after this and then the Showcase one

G-e-n-e-v-e-n-s-i-S:
> pfft i had the exact opposite where i thought it was some overlooked duplicate from just the name.

lol. why is naming things the hardest part of programming

I changed how the filter looks at supertypes, types and subtypes to align it with comp rules terminology. gives the user more granular control. I'll refrain from doing more tweaks so you can actually test it.

G-e-n-e-v-e-n-s-i-S:
sorry i lied couldnt stop myself from tweaking

CajunAvenger:
Detection algorithm that i'm gonna try to fight MSE into making work

1) Loop each card, splitting its types and subtypes, and checking each of its types for the subtype. If it's not present, its saved as custom (if not there already), hard associated with the type if mono type, soft associated with both otherwise

2) check each subtype in our custom array. If hard associated creature, assume creature. Else if hard associated with anything else, assume its that. Else if soft associated with creature, assume creature. this isn't perfect (two cards in canon magic fail) but its pretty solid. main trick is gonna be making this efficient enough to not take forever thanks to the array bugs

CajunAvenger:
Since the new detect script doesn't care about supertypes at all, it doesn't re-run when that is changed. For now I've set it so the script will re-run if you set the Custom Races option to `reset()` which maybe isn't the final implementation but is a neat feature to have.

CajunAvenger:
brainstorming on pip names

stat that tracks specific color pip strings:
- Casting Pip Colors / Pip Colors (Casting)
- Casting Cost Density
- Casting Color Density

stat that tracks cumulative number of color pips:
- Total Pips (Casting)
- Pip Distribution (Cast)
- Cumulative Pips (Cast)
- All Pip Colors (Casting)
- All Color Pips (Casting)

stat that tracks number of pips regardless of color:
- Casting Cost Pips
- Pip Count (Casting)

maybe an arrangement of
- Pip Count (Casting) [stat that tracks number of pips regardless of color]
- Pip Colors (Casting) [stat that tracks specific color pip strings]
- Color Pip Count (Cast) [stat that tracks cumulative number of color pips]


tho maybe Color Pip Count makes more sense as color pip strings ouagh

G-e-n-e-v-e-n-s-i-S:
gravitating towards
 - Pip Count (Casting) [stat that tracks number of pips regardless of color]
 - Pip Colors (Casting) [stat that tracks specific color pip strings] / or maybe simply Pips (Casting Costs)
 - Total Pips (Casting) [stat that tracks cumulative number of color pips]

then there's the problem of ordering:
Pips (Casting Costs)
Pips (Ability Costs)
Pips (All Costs)
Total Pips (Casting)
Total Pips (Ability)
Total Pips (All)
Pip Count (Casting)
Pip Count (Ability)
Pip Count (All)

or
Pips (Casting Cost)
Total Pips (Casting)
Pip Count (Casting)
Pips (Ability Cost)
Total Pips (Ability)
Pip Count (Ability)
Pips (All Costs)
Total Pips (All)
Pip Count (All)

first ordering seems better to me

CajunAvenger:
changes seem fine, really hard to clock mse script but seems to shave ~1-5ms

if we aren't gonna use the specific artifact/enchantment/land type for anything we can slim it a bit more by just collapsing into creature hand and noncreature hard

G-e-n-e-v-e-n-s-i-S:
There's one last change I'd like to make to the filter.

Currently when you add multiple conditions, it checks for
condition 1 AND condition 2 AND etc...
I half-assed in a way to do OR, but I want to make it easier to use, just gotta wrap my mind around how it works

G-e-n-e-v-e-n-s-i-S:
ok I think this works. gonna do some more testing on it but making it live already

CajunAvenger:
i'd be inclined to reduce the white space on the help trace, for me at least it makes things a bit harder to read through

tinkering with filter some, i'd also like to pitch `=+` and `!+` as the include/exclude operators over `=:`

CajunAvenger:
hrm, awkward thing for the detect script is that making it a default means the detection script is running with every keypress made inside a typeline, which does start to get noticeable on bigger set files. we can do a trick where we set the default to "reset()" or whatever the trigger phrase is and it'll load whatever the current is on first startup, but it'll need manually updated or reset() from there

G-e-n-e-v-e-n-s-i-S:
Currently, if one of the conditions in the filter gives an error message, then the card will always be kept.

For example, if I have a tarmogoyf in the set, and I write
power<="3"
the tarmogoyf will always be counted in the stats tab, even if I add
power!="*"

maybe we want to allow the other conditions to filter the card, even in case of an error?

CajunAvenger:
mkay, think that's a wrap on what i want to hit. the color CDA localization is still up in the air but we can come back to that during localization pr if we want to.

are there other filter errors you want to look at? or anything else you want to tinker before we merge?

G-e-n-e-v-e-n-s-i-S:
I think I'm good.

You said you wanted to look at my showcase frames next, in the meantime, do you want me to open a PR with the few localization additions I made?

PR #13, opened 2023-04-22
[ADD] Added M15 styled double faced battle cards

9VZZH7:
This is the implementation of the new Battle cards. However only the style is present in this pull request. Adding the card type (and subtypes) has to be done somewhere else. This is a style template only.


VerumCH:
Played around with this in my own fork. Overall, it seems to mostly work, but there are some issues and a couple limitations that I've personally run into in my testing:

image

#### Issues: 
* When the mana cost has only one symbol, it doesn't fit within the frame of the title bar on the front side.
* The back side does not have the correct boundaries set for the title text box, meaning the name on the back can run over top of the DFC symbol.
* "Chop" top/bottom do not work on the front side. (They seem to work fine on the back side.)
* The "placeholder" card in the style selection list is a generic, single-face grey card, making it impossible to differentiate by image alone. By convention of other styles, it should probably use a thumbnail of a blank Blue DFC Battle. Any color would be fine though as long as it actually showed a DFC Battle.

#### Limitations:
* There is no support on either the front *or* back for any of the font size caps/reductions common on the Mainframe or other styles. (e.g. "font cap" to set a limit on the body text size, often far more useful than setting "chop" values. Or "shrink name/typeline text" to lower the height of these texts for long names, to keep the appearance in line with actual MTG cards with long names.)
* Back side frame color and color indicator must *both* be set manually. These should always match exactly the colors of the front side, unlike some transforming cards or many MDFCs. Would be nice if that could be set automatically based on the mana cost on the front, at least as a default. If some mad scientist wants to make a battle with different-color backside, they could still override it.

CajunAvenger:
i am now working on battles and building the frames myself so more mainfram-ing can be built into them

VerumCH:
Oh nice! Was thinking about trying to add battles to Mainframe DFC myself, but figured you were probably at some stage of working on it. That's good to hear.

PR #18, opened 2023-06-14
Added a few showcase frames

G-e-n-e-v-e-n-s-i-S:
Showcases:
 - Assassin's Creed Memory Corridor
 - Bloomburrow Borderless Anime
 - Bloomburrow Woodland
 - Doctor Who TARDIS
 - Dominaria Stained Glass
 - Eldraine Enchanting Tales
 - Fallout Pip-Boy
 - Gatewatch Baseball
 - Ikoria Crystal
 - Innistrad Double Feature †
 - Innistrad Double Feature Walker †
 - Innistrad Equinox (only single faced)
 - Ixalan Aztec
 - Ixalan Caverns Aztec
 - Ixalan Legends
 - Lord of the Rings - Ring
 - Lord of the Rings - Scroll
 - Lord of the Rings - Scroll Saga
 - New Capenna Art Deco †
 - New Capenna Art Deco Walker †
 - New Capenna Golden Age †
 - New Capenna Skyscraper †
 - Ravnica Citadel
 - Ravnica Dossier
 - Tarkir Dragon
 - Thunder Junction Breaking News (crown from card conjurer)
 - Thunder Junction Vault
 - Thunder Junction Wanted
 - Unfinity Galaxy

Misc:
 - Punch-out Counter Sheet
 - Kamigawa Full Art Basics
 - Lord of the Rings - Mechanics Reminder
 - Unfinity Attraction †

†  based on blanks from Card Conjurer, downscaled and tweaked by myself

PR #21, opened 2023-08-10
card_fields and mse-includes update

G-e-n-e-v-e-n-s-i-S:
Add overwritable custom text fields for template makers.


CajunAvenger:
i'm not entirely convinced by the use case here. in a lot of cases where this can't use an extra card field for some reason, you'd want to know what the field is rather than just "custom 2", and transferring it over to another field later if it becomes "canon" becomes a big headache.

G-e-n-e-v-e-n-s-i-S:
What you say is true, but it assumes that in those situations you can just add a field in the game files.

If you're willing to do that for me then (I know this is going to sound like a troll but) can you add 6 fields named choice 1 to choice 6 that are defined exactly like those in this pull request?

G-e-n-e-v-e-n-s-i-S:
I'm proposing a rewrite of the `card_fields` file. This solves the problem I was having when making this PR, so I'm putting it here.

The solution no longer involves custom text fields, but adding a bunch of functions to the `default` and `script` parameters of fields, so that templates can override them, giving more control to template creators. For example, in Mainframe, the `lvl label X` fields could probably be replaced by overriding the default function of the `level X` fields.

(Unfortunately, Github's diffing algorithm is not very useful when reviewing these changes, since I did a lot of reshuffling of the fields, to group similar ones together.)

- Grouped similar fields together.
- Improved/added descriptions.
- Removed all `icon`, `choice colors` and `position hint` properties. (These were for the stats page. They are no longer used, so let's not waste clock cycles.)
- Changed `shape` field type from `choice` to `text`.
- Changed `script: card.rarity` to `default: card.rarity` in `rarity 2` field
- Added the following functions, so styles can override them if needed:
	- `text_filter_default()` to `rules text`, `text` and `level X text` fields
	- `flavor_text_filter_default()` to `flavor text` fields
	- `chapter_text_filter_default()` to `chapter text` fields
	- `level_filter_default()` and `level_filter()` to `level X` fields
	- `copyright_default()` to `copyright` fields
	- `set_code_default()` to `set code` fields
	- `card_full_name()` to `full name` field
- Changed `margin_code` parameter value of `chapter text` fields from "unknown" to "chapter_text". These codes can be used by the templates to know which field is calling the function, if special behavior needs to be coded for some of them.
- Added the following fields:
	- `name 4`
	- `alias 4`
	- `casting cost 4`
	- `super type 4`
	- `type 4`
	- `type symbol 4`
	- `sub type 4`
	- `rarity 3`
	- `flavor text 4`
	- `text 4`
	- `level 20 text`
	- `copyright b` (for universes beyond additional copyright)
	- `copyright b 2`
	- `copyright b 3`
	- `illustrator 4`
	- `color identity` (to make it sortable, we did this work for the stats page, might as well take advantage of it)
	- `color category` (to make it sortable)
	- `color count` (to make it sortable)
	- `template` (to make it sortable)
- Added the following choices:
	- `battle` and `champion` to `type symbol` fields
	- `compass`, `closed fan`, `open fan`, `bro war`, `lesson`, `specialized`, `triangle`, `triangle inverted`, `aetherprint`, `modal`, `sparker`, `sparker (colored)`, `custom one`, `custom two` and card types to `transformation` fields
	- `heart` to `card stamp` fields
- Added necessary image files for the choices above.
- Changed the `type` field column name to "Full Type". Changed the `type symbol` field column name to "Type" and made it visible by default. This provides a simplified version of the `type` field for sorting purposes. Currently, when we sort by the `type` column, legendary artifacts and nonlegendary artifacts are not grouped together, for example. Sorting by `type symbol` fixes this. Alternatively, we could add a `sort script` to the `type` field, but I'm not sure this would be better.
- For the same reason, changed the `card color` field column name to "Full Color". Changed the `color category` field column name to "Color" and made it visible by default.

CajunAvenger:
looking over copyright:
it looks like we can add extra card fields this way just fine, i'd push for doing that over card fields where relevant, stuff like flavor bars and artist arrows
similarly i'd like to look at segmenting these some so all these fields aren't being loaded for templates that have no way to use them anyway

G-e-n-e-v-e-n-s-i-S:
> it looks like we can add extra card fields this way just fine, i'd push for doing that over card fields where relevant

is there a performance reason to that? it seems to me that card fields are always preferable, since you can read their value, and the value transfers over when the user switches templates.

> similarly i'd like to look at segmenting these some

definitely the way to go

> omai the z index arms race

yeah I just put huge values so there's plenty of space in between. once we have a clearer global picture, we can go back and reel those in.

CajunAvenger:
> is there a performance reason to that?
yes, you eat some performance for each card field, and for each of your extra card fields. if it gets out of hand you can get input lag, it's one of the things that killed the original version of Mainframe that let you do planeswalkers and split cards and masterpiece frames all on a single style, and why mainframes are as economical as they can manage with extra card fields

looking over script rn to get a sense of all the new functions and you actually updated the backwards compatibility flavor bar equation 🫨 that only existed for people too stubborn to use 2.1.2

CajunAvenger:
if you do need to access the data, then yea card field is the way to go, at least until i get to the MSE 2.4 part of my backlog and we can access those

CajunAvenger:
big picture thing, i'd consider moving all the include files that operate this way under one .mse-include, something like magic-framebuilding or magic-modules or something. then we can probably standardize the include files `magic-modules.mse-include/loyalty/script`, `loyalty/styling`, `loyalty/card_field`, `script_complete`, `script_dfc` sort of setup. 

slight con in that versioning the base include file here does become more important, i think this will be an okay trade off in condensing things down to one folder, for knowing where everything is, fewer dependencies to set up, and being easier to move around.

this was going to be a big piece of the 'mainframe rebuild' for 2.0, putting every shared image into an organized magic-mainframe-assets.mse-include file, some of them are already part of this

CajunAvenger:
whichever way we do it, might be nice to define like
`copyright_images := "/magic-copyright.mse-include/"
module_images := "/magic-modules.mse-include/"`

for `image: {copyright_images + "art.png"}`

CajunAvenger:
is there a reason that face_coordinates is a string and not an array? as far as i can tell it's entirely backend so it can do arrays instead of converting like we need to do with style options

G-e-n-e-v-e-n-s-i-S:
> is there a reason that face_coordinates is a string

I think it's because at some point I wanted it's value stored in a card field. I don't remember if that's still relevant

> moving all the include files that operate this way under one .mse-include

That would be easier in the long run, but how would we handle custom watermark images? People have put them in `/magic-watermarks.mse-include/`, and now the program would expect them in `/magic-modules.mse-include/` or `/magic-modules.mse-include/watermarks/`

If people still need to put the images in `/magic-watermarks.mse-include/` then that negates a big part of the gain.

CajunAvenger:
custom stuff should go into magic-mainframe-extras whenever possible, iirc watermarks didn't because it would have required magic-watermarks to be a dependent on magic-mainframe-extras which wasn't something i could justify back then.

watermarks as the already established one being an exception is probably fine, pride pinlines probably would too except i don't think anything outside of mainframe used them so we can do that. 

alternatively i think pulling off the band-aid is absolutely on the table, anyone who was setting up custom watermarks will be able to handle moving them to magic-mainframe-extras, and it means we can set up a gitignore to gracefully leave custom additions to that file alone and let people fully run MSE off the repo

CajunAvenger:
also please hold on committing things until my next, i've got a bit chunk inbound

CajunAvenger:
aight, from here

i'm going to stress test with some big mse files and see if the card fields are too heavy. being able to maintain them across styles and have different sized shapes are nice, but its a pretty small benefit vs hurting render time. one thing that would help is if they don't have scripts/defaults in card_fields, with the scripting happening in choice_images instead; that's something i'll probably look at even if we do keep them

G-e-n-e-v-e-n-s-i-S:
> also please hold on committing things until my next

sorry didn't see it in time, hopefully my commit was small enough

> i think pulling off the band-aid is absolutely on the table

I'm in favor

CajunAvenger:
note for future whoever, readme should say where the default position for the offset fields are on a standard 375x523 card

G-e-n-e-v-e-n-s-i-S:
for watermarks, we can now use

include file:		/magic.mse-game/watermarks/menu_choice_images
image:			{ watermark_scripts[card.watermark](face:1) }

this means that we only have one file to maintain, instead of
watermarks
watermarks_back
watermarks_third
who only differ by the face parameter.

CajunAvenger:
k pleasantly surprised with the stress testing, guess i was just using a real bottom of the line computer back in the day.

some notes from using things live
* i don't like the new type sorting script, it's kinda heavy and not being alphabetical anymore keeps throwing me. seems like we could add a Card Type field that filters out the supertypes instead
* not a fan of the set language moving down when you stack the credit lines, i'd like to see that as optional behavior with a set or styling option switch
* copyright_offset_top, copyright_offset_pt_left, and probably copyright_offset_pt_top should get more accurate names; potentially the whole set to differentiate the copyright box vs the entire "information below the text box" area. def _offset_top should be more clear it's for the entire area and _offset_pt_left that its only the secondary copyright

for now i'm going to turn my attention to card fields, reviewing the new fields and scripts there. on your end, please let me know if there's anything at all in the redraft you have issues with, then converting over the other .mse-includes once we have a system we're happy with. for the moment, keep them in their indivual mse-include files

G-e-n-e-v-e-n-s-i-S:
I see a problem with the `face_coordinates_map` implementation, specifically this:

`fc[input-1] or else fc[0]`

it makes it so frames that have only 1 face are actually considered to have 3 faces stacked on top of each other.

I believe the prior behavior was that it would return an array of zeros

G-e-n-e-v-e-n-s-i-S:
Would a `rarity` include make sense? Or a `complete_typeline` include (rarity + type + indicator)?

If so, maybe also a `complete_nameline` include?

(`complete_template` include when? 👀)

CajunAvenger:
yea rarity include is on my shortlist

complete typeline: i wanna see if my concept for auto-shrinking single lines with work first, probably still worth to standardize color identity dot (which we can probably properly automate at this point instead of needing the style option)

complete nameline: if the autoshrink works yea, if not maybe still yea for the styling option instead. once we get into the full madness i have for spree almost certainly

complete template: 🤔

CajunAvenger:
> once we get into the full madness i have for spree almost certainly

i do want to say, my goal for this PR is to get the basic version of this system out the door so we can use it for #18, further modules and more advanced additions can be saved for that or another PR as part of 1.3

i think the aim here is
* simple nameline package: for now we don't worry about alias or spree or Futureshifted or whatever
* simple typeline package
* the custom fonts package to go with those
* simple chapters package: i am def going to be adding the Saga witchery code here eventually but it can wait
* stamps: this i probably do want to get right while we're here rather than coming back to it
* possible stretch goals while i'm doing that, simple textbox and pt to go with the font package: these both def have complicated parts that we can kick to another PR instead
* transform symbols: for now probably have this in DFC Nameline and figure out the setup for Spree and single-card Lesson symbols Later.
* want to clean up watermarks; we're probably going to need to leave magic-watermarks.mse-include functional but have the primary method from here on being this system. this has the same complications as textbox so it doesn't need to be fully online for this PR
* copyright, corners, counters, loyalty, separators we have, just get them tuned up
* probably fixing the scruffiness i found on meld 3 in 1

Changes that require hacking MSE files (like say, moving popout image from `image 2` to a dedicated `popout image` or renaming style fields now that standardization is so free) *are* on the table, but *will* need to be a later PR so i can get a tool set up on the LackeyBot website to make those changes to MSE files. the bot already has a system for editing MSE zips so this is fully on the table but will be some extra time that i don't want holding up showcase frames

G-e-n-e-v-e-n-s-i-S:
I'll make a draft of the nameline package tomorrow.

G-e-n-e-v-e-n-s-i-S:
devious

G-e-n-e-v-e-n-s-i-S:
I found a problem with this :/
Problem is that it's using the card_style variable, which doesn't initialize on templates that are not the default set template until you click on the card.

Steps to reproduce:
 - create a set with two cards, one using the default set template, the other with a different template that implements this system.
 - make sure the card that implements this system is not the one that is shown when you boot up the program.
 - boot up the program, and without clicking on that card, export all card images.
 - that card's export will not use this system. (specifically, `card_style.name_mirror.content_width` will be zero, and so `mirror_adjust()` will return 1)
 
This means that before you export a set, you must click on all the cards individually, which is unacceptable.
I'll push my implementation here so you can check it out, hopefully I made a blunder.

CajunAvenger:
name shrink, could try the averaging thing instead (10.5 * length(card.name) ~~ card_style.name.content_width) but the multiplier needs to be figured per unique font and card setup and it'd still need a manual override anyway

potentially having the offset as a save value: true field would get around the initialized problem

G-e-n-e-v-e-n-s-i-S:
Oh boy, that brings back fond memories of when I was trying to store the text's length inside a custom tag in the Lorcana templates to get around that same exact bug. (It all ended in a garbage fire that's still raging to this day and that I cannot bare to look I'm so sorry nim)

Anyway, VERY interested if you find a solution :)

G-e-n-e-v-e-n-s-i-S:
Here's essentially what I was doing, if I remember correctly:

```
get_stored_width := filter_text@(match: "[0-9]+", in_context: "<width=<match>") + to_int
```

```
card field:
	type: text
	name: name
	default: "<width=0></width=0>"
	script:
		old_width := get_width(value)
		width :=	if length(remove_tags(value)) == 0 then 0
					else if card_style.name_mirror.content_width > 0 then to_int(card_style.name_mirror.content_width)
					else old_width
		<width=" + width + "></width=" + width + ">" + remove_tags(value)
card field:
	type: text
	name: name mirror
	script: card.name
```

CajunAvenger:
hm okay. gives me some more ideas to explore. let's see if any turn out

CajunAvenger:
losing my mind rn

```
card field:
	type: text
	name: name mirror
	description: A mirror of the name to help resize it
	show statistics: false
	save value: true
	script: apply_margins(remove_tags(card.name), name:"name_mirror")
card field:
	type: text
	name: pls
	show statistics: false
	save value: true
	script: mirror_adjust(mirror:card.name_mirror, maximum:270)
```
neither of these fields save??? they don't even save with a bad value they just refuse to save at all

CajunAvenger:
@ more includes

> my goal for this PR is to get the basic version of this system out the door so we can use it for #18, further modules and more advanced additions can be saved for that or another PR as part of 1.3

trying to get this approved this weekend and hopefully most or all of the way through showcases, then revisit stuff like cards and crowns after. those may even go to 1.4 where we get battles and the saga PR in

G-e-n-e-v-e-n-s-i-S:
Yeah sorry, I don't really know what to do now, so I'm just adding images, as these don't really need to be reviewed.

You said a rarity include was on your short list. That can be useful for the showcase frames, but other than that, I believe they have everything they need.

CajunAvenger:
> Yeah sorry, I don't really know what to do now, so I'm just adding images, as these don't really need to be reviewed.

you can open them in another PR, we do need to get to them but i wanna put a solid close line on this rather than let it expand forever and risk letting it keep remaining unfinished

and i will very much be reviewing those images :v the background images should either be full scale textures or cards that are masked, rather than precut

CajunAvenger:
corners really shouldn't be above overlay stuff, and can be mixed into an updated border handling?

G-e-n-e-v-e-n-s-i-S:
I'm thinking about corners in a physical way. They represent an absence of cardboard, so in that sense, nothing can go over them.

Also, completely decoupling them from borders is a massive simplification. They can be specified at the set level (which is generally what you want, either all cards have corners or none do), and all we need to do is add the `include file:` line in older templates without having to worry about how that template's border works, which can sometimes be complicated.

CajunAvenger:
new sorting is nice, too bad its such a pain to decache the old numbering :pensive:

not sure separators need the set.use_flavor bar call built in. presumably the point of having 8 with a generic name is they aren't all flavor bars?

want to rename `copyright b` something like `secondary copyright` rather than b

i don't see the value in card.position; that's easily obtainable by scripts and the default sorting is already by position. appears to be unused as well

similarly not sure if card.face_coordinates does anything

color category, exact color, color identity, color count: guess these can stay if card field bloat isn't a problem anymore. should probably get added to custom_index though

CajunAvenger:
next up i'm gonna do stamps

for some stuff you can work on if you're itching
* copyright b rename
* copyright_offset_top, copyright_offset_pt_left, and probably copyright_offset_pt_top should get more accurate names; potentially the whole set to differentiate the copyright box vs the entire "information below the text box" area. def _offset_top should be more clear it's for the entire area and _offset_pt_left that its only the secondary copyright
* i'd like to see set language moving down when you stack the credit lines as optional behavior with a set or styling option switch
* copyright_offset_top, copyright_offset_pt_left, and probably copyright_offset_pt_top should get more accurate names; potentially the whole set to differentiate the copyright box vs the entire "information below the text box" area. def _offset_top should be more clear it's for the entire area and _offset_pt_left that its only the secondary copyright
* `_complete` should split into a front face and _dfc version of the file (which `include files:` the front face)
* potentially scripts that aren't used in `card_fields` should get moved from `magic.mse-game/script` to `magic-name.mse-include/script`; i'm not sure how MSE does script caching it might actually be better to leave those in script, performance wise. alt if most of these actually are needed in card fields then its probably better to keep them together even if it means they're in a less convenient place. possibly time to look into breaking up script into multiple files
* z index reassigning

G-e-n-e-v-e-n-s-i-S:
>     not sure separators need the set.use_flavor bar call built in. presumably the point of having 8 with a generic name is they aren't all flavor bars?

yeah, the plan was that they could be used for class levels, case abilities, etc.... not sure how realistic that is though.

>     * i'd like to see set language moving down when you stack the credit lines as optional behavior with a set or styling option switch

I'll get on that

>     partition select added to the copyright include, it's a box over the automated card number

just add the partition field over the automated number? should it get disabled by anything or always there?

>     * `_complete` files should probably get split more-or-less by field

Yeah, my idea was that you would have a rarity include, a color indicator dot include, then the typeline meta-include that groups them. If the `typeline_offset` functions are defined in the game script file, then the rarity include can reference them, but they can be ignored if you only use the rarity. I don't think defining a lot of variables like that in the script file has any noticeable impact on performance.

>     i don't see the value in card.position

I had added it when I read this post:
https://magicseteditor.boards.net/thread/2494/script-find-card-numbering-cards

apparently when this person was trying to access `position(of: card, in: set)` it would crash MSE on startup (although it might have been something else they did)

>     similarly not sure if card.face_coordinates does anything

yeah, probably axe it

CajunAvenger:
Done:
* Reviewed blends
* Reviewed module system
* Reviewed stamps
* Reviewed chapters, corners

TODO
* Review "copyright", renamed "information"
* Review counters, loyalty
* Review nameline
* Review typeline
* Port pride pinlines
* Figure out how we're handling identity
* Figure out how we're handling watermarks
* add new fields to custom_index
* fix meld 3 in 1 sizing wibble
* reassign z indexes
* remove test files

G-e-n-e-v-e-n-s-i-S:
For watermarks, I'm a strong proponent of ripping the band-aid now.

For identities, we can move them into modules later. It's gonna take a while to thoroughly modify all existing templates, so it's probably best if we do that in it's own PR.

CajunAvenger:
i think my plan is to move everything in the repo over, since we can just do that with a find-replace, but leave the old one in place for awhile to give some breathing room for outside templates to switch.

identity-new might just stay indefinitely cause its small but also a pain to update en masse

CajunAvenger:
> abilities_count() >= 5 

k this part of loyalty include doesn't play nice with dfcs, this'll either need retooled or this one'll need more a more dedicated dfc alternate include

CajunAvenger:
> with 2.4.0 we can rename set.blend_with_colors to set.default_watermak_render and change it's choices without loss of backwards compatibility

i was thinking of just adding a second switch for the built-in watermarks rather than tying it to what was just meant to be a switch for "will this watermark break if i try to blend it"

G-e-n-e-v-e-n-s-i-S:
> i was thinking of just adding a second switch for the built-in watermarks

Sure, I'll let you see how you want to implement it

> pride pinlines

I'm planning on adding a normal pinline include in another PR (#48), so maybe the pride pinline functions should be called `pride_pinline_offset_` instead of `pinline_offset_`

CajunAvenger:
ah crap, the single-line shrink effect applies to casting cost, even without scale down to unlike normal text fields, and seems to be ignoring it entirely. gonna tinker a little bit but will probably have to pull that back for now

G-e-n-e-v-e-n-s-i-S:
Working on the loyalty DFC. Should have it done by tomorrow, however it's a mess and I don't expect it to be useful for Mainframe DFC, nor do I think it's a good idea to make it so.

CajunAvenger:
k if  that just needs to be something i build later we can do that

CajunAvenger:
idk if there's anything to be done about this, but corners being the highest z-index means popout art can't get fully into the squared corners option
image

G-e-n-e-v-e-n-s-i-S:
> idk if there's anything to be done about this, but corners being the highest z-index means popout art can't get fully into the squared corners option

yeah I have a plan for this. basically I'm gonna do what I did in this PR #44 but on ALL the templates, then we can change the `corners_image` function to use `rgba(0,0,0,0)` instead of `card.border_color` when `shape == "none"`. this is the most time consuming, but also the best solution.

CajunAvenger:
idle thought, we could now compile MSE as like `Version 2.1.3b` `Version 2.1.3za` to incorporate template pack version into that, given how many people only look at that. may be fine to just keep the one in script up to date though

CajunAvenger:
> yeah I have a plan for this. basically I'm gonna do what I did in this PR #44 but on ALL the templates, then we can change the `corners_image` function to use `rgba(0,0,0,0)` instead of `card.border_color` when `shape == "none"`. this is the most time consuming, but also the best solution.

Could this just have the rgba(0,0,0,0) in as an extra option, at least in the interim?

G-e-n-e-v-e-n-s-i-S:
> Could this just have the rgba(0,0,0,0) in as an extra option, at least in the interim?

of course

CajunAvenger:
losing my mind rn. having a `scale down to` on single line fields reduces the font size two points, a bug even present on 2.1.2 where scale down to doesn't work

CajunAvenger:
why are the transform symbol left and top different for horizontal cards? is this a rotation thing

CajunAvenger:
name and type's `left` is offset by nameline and typeline's `offset_top` (in addition to `offset_left`, i suspect it should not?

CajunAvenger:
reposting since this got buried

TODO
* review rarity dimenions
* review watermark dimensions
* review card symbol offset
* reassign z indexes

* Loyalty finishing
* potentially errant _top_offset in type and name's left/right
* invisible corner temp option
* Genevensis review of card symbols
* remove test files

FUTURE PRs
* modules/pinlines and the rest
* update magic-watermarks and magic-identity usage in the pack
* make other mana fonts alch_compatible
* MSE 2.1.3

G-e-n-e-v-e-n-s-i-S:
> name and type's left is offset by nameline and typeline's offset_top (in addition to offset_left, i suspect it should not?

I have it set up so that `nameline_offset_top` also increases the transformation symbol height and width, so to accommodate it pushes the name to the right. this may or may not be behavior that we want. regardless, currently it's not checking if the symbol is disabled which is a bug. I'll fix it.

also rarity seems broken now, it's looking for `rarity_left` which doesn't exist

G-e-n-e-v-e-n-s-i-S:
> why are the transform symbol left and top different for horizontal cards? is this a rotation thing

it assumes rotated cards are battles

PR #23, opened 2023-11-10
Localization Update

G-e-n-e-v-e-n-s-i-S:


G-e-n-e-v-e-n-s-i-S:
Incomplete things in language map:
 - is_kindred
 - is_nonstandard
 - is_hero
 - is_phenomenon
 - is_scheme
 - is_vanguard
 - is_elite
 - is_host
 - is_ongoing
 - is_world
 - is_wastes
 - get_supertypes
 - remove_supertypes
 - base_supertypes
 - mana_symbol_production
 - gold_mana_production
 - chosen_mana_production

Other incomplete things:
 - keyword files only go up to Phyrexia: All Will Be One, and are a bit janky, (except for French) but need native speakers to be ironed out.
 - magic-word files need to be generated for German, Spanish, and Korean.
 - magic-word files could use a refresh for French, Italian, Russian and Portuguese.
 - locale needs to be generated for Korean.
 - locales need to be updated for German, Italian, Russian Portuguese, Japanese, Chinese simplified and Chinese traditional.
 - word lists
 - stats page localization
 - template localization

CajunAvenger:
k added in a whole tonna words

todo

* improve my clumper script for splitting up really big lists and writing the word_lists
* tune up Russian/Korean/Japanese alphabetizing. Chinese also needs done but i do not know where to start there
* once those are sorted, each language gets its own word list for Card Type, Creature Races, and Creature Races. the rest can share just fine.

G-e-n-e-v-e-n-s-i-S:
Wizards have been seen to fallback on English when they couldn't be bothered to come up with translations...
ru_walker

CajunAvenger:
Planeswalker is a weird one cause its a trademark thing, some of the early planeswalkers did have it translated and it turned into "Planeswalker" later

G-e-n-e-v-e-n-s-i-S:
Couple things I noticed:

1) When I select the card language from the set info tab, in the drop down menu, the flags are all mixed up. Once the language is selected, it shows the correct flag, but specifically when the drop down menu is open, some flags are not paired with the correct language. Not sure this is within our control to fix, might be an exe problem.

2) The following files do not seem to do anything:
 - In the `magic.mse-game` folder, all the magic-words files. I think MSE checks for those in `magic.mse-game/dictionary/`
 - In `magic.mse-game/dictionary/`, all the files that do not end with `.dic` also seem like unnecessary repeats

CajunAvenger:
> When I select the card language from the set info tab, in the drop down menu, the flags are all mixed up

that's cause the images are cached via index for some reason, if you clear cache (or select the language) the flag will correct itself

> redundant files

good to know


did the French word_list have some code i missed/bullldozed/never had that made the dropdown selector work properly? i thought i remember that working, just sub-optimally, in the past, and its currently looking like its gonna be a doozy to set up

G-e-n-e-v-e-n-s-i-S:
French, Spanish, Italian and Portuguese have the supertypes after the types, so we can't use the "is prefix" check marks.
We have to settle on a list of common combined types+supertypes, and include those. When I did French I had chosen these:
Instant
Sorcery
Creature
Legendary Creature
Artifact
Legendary Artifact
Enchantment
Legendary Enchantment
Land
Legendary Land
Basic Land
Snow Land
Legendary Planeswalker
Artifact Creature
Legendary Artifact Creature
Enchantment Creature
Legendary Enchantment Creature
Token Creature

Here are these lists localized:
[word_lists.txt](https://github.com/MagicSetEditorPacks/Full-Magic-Pack/files/13377487/word_lists.txt)

CajunAvenger:
mkay, def the easiest way to set things up

i had looked at making all the options prefixes, to basically make it a text field you can type in, but that bugs out deleting words in ways i don't understand so that doesn't seem feasible

a neat but maybe too much scripting thing i'd gotten to was having super_type_filter arrange the words with get_supertypes+remove_supertypes checks, which would let us auto fix gendered words too. that's got some weirdness to solve with custom types but i'm not ruling that out just yet

another option i haven't tested yet would be to set up the supertype dropdown like the subtype dropdown, where there's a second one off to the side to easily add another one. this causes subtypes to end in a pretty harmless [space] that's more troublesome on supertypes though, albeit probably resolveable by making it a nonbreaking space?

CajunAvenger:
k, it looks like the prefix problem is that word_list aggressively removes custom types when you select an option, which pinches us in to common list or creating more fine tuned list options per language

CajunAvenger:
hm possible issue on this with backwards compatibility. probably something we could work into the filter though

G-e-n-e-v-e-n-s-i-S:
The supertype drop down is very hard to see and click on, on a 1080p monitor.

Isn't the word-list deleting custom types also a problem on the live version?

CajunAvenger:
> Isn't the word-list deleting custom types also a problem on the live version?
> I say give the word rearranging script a go.

it is a bit of a problem, but on live the issue comes from having like `Foo Creature` that you try to add Legendary to, which becomes `Legendary Creature`

the script means if you input `Créature Artefact`, it's corrected to `Créature-artefact`, then trying to add légendaire to it replaces the whole thing with `légendaire`

maybe if we run a script to collect the custom supertypes, like it does for custom creature types, might be a workaround that doesn't involve having to add every legal combination into the list via submenus

CajunAvenger:
> maybe if we run a script to collect the custom supertypes, like it does for custom creature types, might be a workaround that doesn't involve having to add every legal combination into the list via submenus

asdf this causes MSE to crash. some kind of exe level problem with prefixed wordlists. i did find that you can do psuedo-suffixes by having the prefixes at the end of the list, but this seems to be unintentional as it has the same issue where Créature légendaire isn't recognized as legal and gets deleted

we're so near a clean solution but probably just going to have to do the common list option

CajunAvenger:
```word list:
+Instant
+Sorcery
+Creature
+Legendary Creature
+Land
+Basic Land
+Artifact
+Enchantment
+Legendary Planeswalker
+Battle
+Tokens
+--+Token Creature
+--+Token Artifact
+--+Token Enchantment
+--+Token Artifact Creature
+--+Token Enchantment Creature
+Multityped
+--+Artifact Creature
+--+Enchantment Creature
+--+Land Creature
+--+Artifact Land
+--+Enchantment Land
+--+Artifact Enchantment
+Legendary
+--+Legendary Creature
+--+Legendary Land
+--+Legendary Artifact
+--+Legendary Enchantment
+--+Legendary Planeswalker
+--+Legendary Artifact Creature
+--+Legendary Enchantment Creature
+--+Legendary Land Creature
+--+Legendary Artifact Land
+--+Legendary Enchantment Land
+--+Legendary Artifact Enchantment
+Snow
+--+Snow Creature
+--+Snow Land
+--+Basic Snow Land
+--+Snow Artifact
+--+Snow Enchantment
+--+Snow Artifact Creature
+--+Legendary Snow Creature
+--+Legendary Snow Land
+--+Legendary Snow Artifact
+--+Legendary Snow Enchantment
+Nonstandard
+--+Conspiracy

G-e-n-e-v-e-n-s-i-S:
My notepad tabs rn
tabs

CajunAvenger:
stretch goal to add to this, find/make glyphs for Russian to stitch into the Beleren font, since those are just getting boxed out currently

G-e-n-e-v-e-n-s-i-S:
Maybe like this? (Hastily put together, but if it's what you had in mind, I can polish it up)
[Beleren-Minion-Bold.zip](https://github.com/MagicSetEditorPacks/Full-Magic-Pack/files/13402131/Beleren-Minion-Bold.zip)

CajunAvenger:
> Maybe like this?

yea that looks pretty good, noice

CajunAvenger:
...bleh using the prefix trick throws off the typeline when you're typing. wordlists are just such a pain

CajunAvenger:
Totem armor we can just add to the auto errata script rather than having the old version

tho having an `Umbra armour` match for British and Canadian English might be a thing to do

CajunAvenger:
> These files do not overwrite the old ones when installing? Not sure why, I haven't touched any of the meta-info. Anyway, if the old files are installed, they need to be uninstalled first before these are installed.

installing without uninstalling works fine for me vOv

G-e-n-e-v-e-n-s-i-S:
> Totem armor we can just add to the auto errata script

right, forgot about the auto errata script

> installing without uninstalling works fine for me vOv

huh, ok

What's the status on the stats page localization?

We should probably automatically disable spellchecking in the text filter function, when the language is Japanese, Chinese simplified or Chinese traditional, since it always results in everything being underlined.

CajunAvenger:
TODO
* magic word files
  * German, Spanish, Korean need made; French, Italian, Russian, Portuguese could use updating
    * not sure how much we need to do here, maybe just load it up with the subtypes we have, this is really more suited for a native speaker to add to for their benefit
* locale files
  * Korean needs made, everything else other than French/English need updated
    * probably this one we should tackle, can use the ja/cn ones to help guide the hotkey stuff for korean
* stats page localization
  * this seems like it'd just be a similar thing that keywords are doing? _lang files with the titles changed, and the help trace which seems rough
* script updates
  * add Totem armor to the auto errata
  * auto-disable spellcheck on ja/cn
* stretch goals, we can merge without these and deal with them later
  * get Chinese subtypes a proper collation system
  * get feature-complete non-English craft reminder text

G-e-n-e-v-e-n-s-i-S:
oh shit the stats help trace kill me now
well, given that I speak french, i'll fully localize the stats page in french to get an idea of what it takes/how it's gonna work

2.4.0 would allow us to localize the style and set info tabs as well (at least the field names, not the field values though)
alternatively we could give the entire `magic.mse-game` folder the keywords file treatment, but are we willing to go that far?

CajunAvenger:
nah we'll hit that higher level localization when i get to MSE 2.2+ testing. hagan has a pr for a lot of that already it just upsets MSE 2.1.2 and the last time i tried to use MSE 2.4 it broke mse on my computer entirely x.x

G-e-n-e-v-e-n-s-i-S:
turns out russian words can have an absurd number of different endings. not sure how to handle these cause checking them all is basically equivalent to .*

CajunAvenger:
re:russian endings

i'm actually not sure we need the monocolor CDAs at all anymore? the thing you've got set up now actually only gets false positives on canon cards now that monocolor CDAs are all color indicators. if we do keep it, should probably narrow it down by having it check for the end of the Cardname atom and then \w* is basically fine

G-e-n-e-v-e-n-s-i-S:
Yeah probably over engineering that CDA thing, feel free to remove it. The monocolored CDAs should only be used on a string found after the name of the card though?

I have pretty much localized the entire `statistics_script` file. Just one minor thing I didn't do: the `triggered_cost_filter_statistic` function. (Oh and also the help trace but that's gonna have to wait.)

Now we need to localize the `statistics` file. I started doing it for French. What I didn't yet localize, is the names of the statistics dimensions (which I will take care of myself for French), and the colors for the subtypes (starting at line 258), but need to write a script to do those automatically.

Basically, the `group:` and `colors:` entries need to be replaced with the values I added in the `language_map`, their spaces need to be replaced with EN spaces, all while not deleting the zero width spaces, for each language.

I did all that in the `statistics` file itself so you can have the diff. The old `statistics` file I have moved to `statistics_en` temporarily.

I HAVE NOT TESTED ANY OF THIS. It's extremely tedious work, and I need to get some sleep, then do something else for a bit. I'm committing it so you can take a look at it, maybe pick up where I left off and probably stop me from over engineering half of it.

G-e-n-e-v-e-n-s-i-S:
UP NEXT:
 - Currently, the way subtypes, types and supertypes are cleaned up to get counted by the stats is assuming they are always separated by spaces. Need to make that adapt to lang_setting("subtype_separator") etc..
 - `statistics` file is pretty much done for French, need to make the ones for the other languages (I have a script that does the noncreature subtype colors automatically)
  - Localize the stats help trace at least in French
 - Tons of testing
 
NEW STRETCH GOALS:
- Get hagan to double the width of the freaking stats dimension names column. It's literally impossible to fit French names in there. Will most likely have the same problem with other Latin languages. Localizing these names is pretty independent of everything else though, so it can wait.
 - Either localize the filter keys (by adding entries in the alias map), or add another help trace that lists most possible keys, with a localized description for each.
 - Localize text_filter so it automatically transforms mana into symbols.

G-e-n-e-v-e-n-s-i-S:
Alright, I think I finished the stats page. I did almost no testing though.

I did find that in french, when you create a legendary creature, something is adding zero width spaces after each type and supertype. Is this something you did?

CajunAvenger:
zero width spaces are used to add leading spaces to a few places in the dropdown menu. we might be able to use a script instead if those are causing problems?

G-e-n-e-v-e-n-s-i-S:
Nothing insurmountable.

Okay well I'm gonna do some bug hunting tomorrow, and then I believe that's going to be it from me for this PR.

G-e-n-e-v-e-n-s-i-S:
are magic `magic-words` files cached or something? for some reason adding words to them was not absolving them from error spelling.

CajunAvenger:
k with that i'm good to merge whenever you're happy with the stats side of it

G-e-n-e-v-e-n-s-i-S:
What's left for the stats page, is translate the names of the dimensions, as well as both help traces (`stats_info` and `filter_keys_info`), in languages other than English and French. However these need to be done by native speakers, since they involve a lot of non-magic terms.

As such I believe this will do for now

PR #28, opened 2024-01-10
Removed outdated templates

G-e-n-e-v-e-n-s-i-S:
Removed 16 templates which can all be achieved with mainframe templates. (May want to double check.)
Here they are bundled if you want to load them in a basic install to do some tests, or add them to the outdated repository.

https://drive.google.com/file/d/13V1oGgVy2a6iSqvxqFRmzi1WNN1DlYht/view?usp=sharing

magic-m15-clear.mse-style
magic-m15-custom.mse-style
magic-m15-leveler.mse-style
magic-m15-metropolis.mse-style (I don't understand what this one does)
magic-m15-mined.mse-style
magic-m15-mined-planeshifted.mse-style
magic-m15-miracle.mse-style
magic-m15-nyx.mse-style
magic-m15-planeshifted.mse-style
magic-m15-planeshifted-leveler.mse-style
magic-m15-planeswalker.mse-style
magic-m15-token.mse-style
magic-m15-ttk-conspiracy.mse-style
magic-m15-ttk-frames.mse-style
magic-m15-vehicles.mse-style
magic-vehicles.mse-style


CajunAvenger:
i was more hesistant to remove some of the more common ones for backwards compatibility sake but i guess the point of outdated templates pack is to handle that

CajunAvenger:
Metropolis changes the set code font; easy cut
M15 Custom appears to be a copy of M15 Altered 1.1 for some reason
the two Mined can def go since Mainframe mana exists yea, may put the mined symbols into mainframe-extras as an example set
Vehicle is an auto frame so that's safe to cut
Token is a strict downgrade on Mainframe Tokens that's safe to cut

Clear: Not an autoframe, but top of the Mainframe list. this is probably okay to cut
Nyx, Conspiracy: isn't actually a mainframe autoframe but probably should be


Miracle, Planeshifted, Planeshifted Leveler: not autoframes so decent contenders for staying, but maybe these are rare enough that it can go with the others and save us a bit of space

Draft-matters: the weird guy. can't autoframe but strange to have without conspiracy. think i will leave this and add the conspiracy scrolls as an autoframe
Leveler: Mainframe levelers cause people a fair amount of problems so this is probably fine to keep; i'll update it to TriLeveler though.
Planeswalker: hrm. I'm keeping this for now if for no other reason than mainframe planeswalkers doesn't work properly on 2.0.1. if i get back to making it backwards compatible i'll revisit removing this

PR #29, opened 2024-01-11
Stats page bug fixes

G-e-n-e-v-e-n-s-i-S:
- Subtypes with spaces inside them are counted as multiple different subtypes. Tried to fix this by adding `<atom-sep>` to all `subtype_separator`s, and then only splitting `card.sub_types` on this tag. That works when creating new cards, but doesn't for existing cards, so I also split on `<word-list>` to accommodate for those. I don't love this implementation though, can probably do better.

- Fixed the colorless value in `Color Category` not being localized and not linking relevant cards when clicking on it's histogram bar.

- Fixed the `global_variable_average_opening_hand_statistic` function. For some reason, creating an array of cards and iterating over it was EXTREMELY slow, and was crashing MSE when the card count approached 3500. (Probably causing a cascade of errors/exceptions in the exe). Iterating over `set.cards` directly instead fixes the problem. Did some minor optimizations as well. Sets should load faster, especially big ones.

- Added `is_tokenish` method that checks if something is a token, an emblem, a designation, a rulestip, a counter or a checklist.


CajunAvenger:
> Fixed the global_variable_average_opening_hand_statistic function. For some reason, creating an array of cards and iterating over it was EXTREMELY slow, and was crashing MSE when the card count approached 3500. (Probably causing a cascade of errors/exceptions in the exe). Iterating over set.cards directly instead fixes the problem. Did some minor optimizations as well. Sets should load faster, especially big ones.

maybe we should just disable this on larger sets that are unlikely to be decks and is just running a heavy startup script for nothing, like `length(set.cards) < 1 or length(set.cards) > 150 then ""` or so

G-e-n-e-v-e-n-s-i-S:
oh, it was specifically the iteration of over a custom built array of cards that was very slow (1.5 minutes on a 3000 card set). now that I've re-written it, it takes basically no time at all. (3000 card sets load exactly as fast whether the script is disabled or not.)

PR #30, opened 2024-01-16
Attempt at fixing ticket, tap and untap symbols

G-e-n-e-v-e-n-s-i-S:
Not sure I understand the situation correctly, but here is what I gather:

Having added `[TK]` to the `mana_sort` function allows it to return `T`, which in turn allows the `mana_filter` function to return `T` as well. Changing `[TK]` to `compound(TK)` seems to work (but TK can no longer be written manually, it must be pasted wholesale or inserted via the insert symbol menu, since `mana_sort` won't allow T or K individually).

The `mana_filter_t` function is implicitly passed an `in_context: mana_context` argument when called within the `text_filter` function. This was preventing it from returning a T or Q. I believe in all cases it's correct to override that to the empty string.

So `tap_reduction(tap_filter())` was not letting through T when it should, and `mana_filter()` was letting T through when it shouldn't. These two wrongs together made a right where it allowed `T:`, but it did not work in the case of `Q:` since "TK" only contains the letter T and not the letter Q.

### DIFF
--- a/data/magic-genevensis-00-main.mse-style/style
+++ b/data/magic-genevensis-00-main.mse-style/style
@@ -43,10 +43,10 @@ init script:
 	
 	mana_t := { "new" }
 	
-	mana_sort :=		sort_text@(order: "\\?XYZI[0123456789]VLHFDSCAIEcompound(TK)P(WUBRG)")
-	mana_sort_wedge :=	sort_text@(order: "\\?XYZI[0123456789]VLHFDSCAIEcompound(TK)P(WBGUR)")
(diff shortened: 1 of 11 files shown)


CajunAvenger:
i think we just need to pop TK out like TQ is popped out, testing that

CajunAvenger:
then again maybe mse is an ancient evil
image

CajunAvenger:
maybe that? it doesn't let tickets and mana mix which is mildly annoying but shouldn't happen anyway, and seems to have stopped breaking

G-e-n-e-v-e-n-s-i-S:
Textbox problems seem fixed.
Yeah mildly annoying that you can't add tickets to casting costs, but we can fix when someone complains :)

Issue #37, opened 2024-02-28
8E tokens and emblems: tracking of name text is messed up


halian:
The names of cards that use the 8E token and emblem frames start out very scrunched up, as if they had negative tracking (spacing between letters); as I switch between such cards, the tracking increases monotonically until the name box hits the maximum width for its template.

![image](https://github.com/MagicSetEditorPacks/Full-Magic-Pack/assets/19718484/0368ff05-913b-4eba-9d0e-9dca81900074)
![image](https://github.com/MagicSetEditorPacks/Full-Magic-Pack/assets/19718484/f3907b0a-e911-47bb-b80c-4c96e268a5f5)

PR #38, opened 2024-03-02
replace 8ED justify with wider horizontal

CajunAvenger:
attempted fix for #37, is that working on your end @halian?


halian:
Works well, except for long names, which overrun the name box. (Also there's no ê, but /shrug.)

image
image
image
image

CajunAvenger:
> Also there's no ê, but /shrug

have updated the formulas to hopefully be better for long names, and have added ê (and è and é and ë) for the vowels.

PR #54, opened 2024-06-11
Gen modules integration

G-e-n-e-v-e-n-s-i-S:
add modules to genevensis custom frames
add `to_number_lax` function in script file
use `to_number_lax` function inside `swap_font_color` function, while parsing color in font system
add 1500x1046 corner masks


CajunAvenger:
this is no longer allowing me to pull this pr due to a permissions issue?

G-e-n-e-v-e-n-s-i-S:
I didn't do anything special with this?
Can you show a screenshot of the problem?

CajunAvenger:
> Couldn't find branch 'gen-modules-integration' in remote 'github-desktop-G-e-n-e-v-e-n-s-i-S'. A common reason for this is that the PR author has deleted their branch or their forked repository.

weird, last night it also had `or you don't have permission to access` in there

G-e-n-e-v-e-n-s-i-S:
I am able to push and pull from this repo....

Although, your error message says "in remote github-desktop-G-e-n-e-v-e-n-s-i-S". There's a "desktop" in there which seems to indicate that it's looking for my local copy? I just deleted the local copy so there's no ambiguity. Try again and if that doesn't work, I'll delete this and re-create it.

G-e-n-e-v-e-n-s-i-S:
Yeah it's the github desktop program, but I'm wondering why it's involved at all. It should just fetch the repo from the github site directly.

Anyway, if it's able to start, then yeah probably an internet issue.

So what's your idea to combine the regular and HD frames? I was thinking maybe an elegant solution would be to just have the exe render the preview in the program while taking DPI into account. So like 300DPI frames get an implicit 0.5x factor to their zoom, etc... That way we can just leave the highest res version of each frame.

After that I can write a program that takes a folder full of MSE renders and can resize them all to the same dimensions, split dfcs, rotate battles, add bleed margins, and/or produce nice pdfs. (I've actually already written most of this.)

Of course that implies we can modify the exe, but if that all sounds good I can start doing some digging.

CajunAvenger:
i have a working demo but it turned out it was on the old version of the frame. 

we start with the 375 frame, put as much as the frame as we can on modules. we add that same stylesheet.card_width scaling to everything else the template itself has to define.

then you can include file and overwrite the size of the card

```
icon: card_sample.png
card width: 375
card height: 523
include file: /magic-modules.mse-include/adaptive_canvas
```
/magic-modules.mse-include/adaptive_canvas:
```
icon: card_sample.png
card width: 375
card height: 523

#### These following properties are the ones being used. Switch the blocks to change the card size

icon: card_sample_hd.png
card width: 744
card height: 1039```

then both versions are the same frame, so you can switch between them without losing styling data or maintaining two versions of the frame. this one file change can apply to everything, and we should be able to disable tracking changes for that file so Github will download the file, but not try to change it back to default when updates come through.

ideally we'd make the big frame 750x1046 so the scale is a clean 2x and we can be sure rounding doesn't screw with things, but if that's a hassle we can probably manage. i can't properly test the big frame to make sure everything is in line though.

G-e-n-e-v-e-n-s-i-S:
wait... I think I just managed to do what I was talking about...
[magicseteditor.zip](https://github.com/user-attachments/files/16972457/magicseteditor.zip)

PR #59, opened 2024-07-28
Unique Card ID

G-e-n-e-v-e-n-s-i-S:
Attempt at creating a unique card id field.

The field should have the following properties:
 - It's value is always an integer.
 - No two cards in a set can ever have the same value.
 - Once a card has been added to the set, its value can never change.

The id can probably be made more unique by integrating `card.time_created`, but it's unclear if that would be useful.

### DIFF
--- a/data/magic.mse-game/card_fields
+++ b/data/magic.mse-game/card_fields
@@ -2981,3 +2981,26 @@ card field:
 	card list alignment: right
 	card list width: 50
 	script: custom_index()
+
+card field:
+	type: text
+	name: id compulse save
+	description: Absolute voodoo shit that seems to force MSE to save the card.id field even when in its default state.
+	save value: true
+	show statistics: false
+	card list allow: false
+	script: forward_editor(prefix: position(of: card, in: set), field: card.id)
+
+card field:
+	type: text
+	name: id
+	description: A unique ID for the card. No two cards in the set will have the same ID. Adding or deleting a card to the set will not change another card's ID.
+	show statistics: false
+	save value: true
+	editable: false
+	card list visible: true
+	card list name: ID
+	card list column: 9998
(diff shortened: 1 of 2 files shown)


CajunAvenger:
if we do this we might as well do a UUID system. ten randomly chosen characters 0-9a-z gets you 3.6 quadrillion possibilities, 0.0001% chance that any two cards in a million share the same id

CajunAvenger:
if we don't care about the user typing it, we could just do actual uuids with 32 hexadecimals

25 0-9a-z is about twice as numerous as uuid, 21-22 in 64bit will get us on either side of it.

hm, maybe do like, save a proper uuid (32 random hexadecimal characters), but the first 8 characters are good enough to be used as an identifier within a set context (0.0001% chance of overlap per 1000 cards, and we can have the checker adapt if more do need to be given, just tell people 8 is good number to hit)

G-e-n-e-v-e-n-s-i-S:
This would be used to link two faces of a DFC in the exporters, so we would care about users typing it. Truncated uuid could be the way to go. (We can always add a uniqueness check like my current implementation does, and change the uuid when the card is added to the set in case of conflict. Changing the id as a card moves to another set kind of defeats the purpose of uuid, but if it only happens 0.0001% of the time that's fine.)

Currently though, the thing that needs testing is the compulsive save trick I'm doing. It makes absolutely no sense, and if it doesn't work, I'm not sure unique ids are even possible. My limited testing seems to indicate that it works, but it might not on other environments or in weird special cases.

G-e-n-e-v-e-n-s-i-S:
Fair.

I'll close this for now, but keep experimenting. (The compulse save hackery thing would actually be super useful for other stuff I have been wanting to do.)

PR #61, opened 2024-08-12
Saga, Class, Leveler, Battle modules

G-e-n-e-v-e-n-s-i-S:
Depends on #48

- added saga textboxes module
- added class textboxes module
- added leveler textboxes module
- rewrote classicshifted template family by iDerp so it uses modules
( originals found here: https://www.reddit.com/r/mtgcube/comments/ihxycy/classicshifted_remastered_a_whopping_update_to_my/ )
- added class, saga, leveler, battle and split variants to test new modules
- added `is_case` language map function (Korean and Russian done with google translate)
- changed `saga_reminder` function so it tries to guess how many chapters there are

TODO:
- add clear frame and transform notch on adventure
- fix card symbol module on battles
- fix interactions between card symbol module and transform symbol module
- explore other possible saga textboxes module implementations
- explore if battle and split modules make sense
- downscale from 1200DPI to 300DPI
- tests


G-e-n-e-v-e-n-s-i-S:
Splitting into multiple PRs for each module

PR #65, opened 2024-09-18
Kasu customs

G-e-n-e-v-e-n-s-i-S:


G-e-n-e-v-e-n-s-i-S:
Right, this actually depends on #61, which itself depends on #48.
I made so many of these I get lost....

It's not showing an exe change for me though. If it is, then that's a mistake.

CajunAvenger:
are these template names locked in? we seem conflicted on if these are "kasu" or "kazu"

G-e-n-e-v-e-n-s-i-S:
oh good catch, it's kasu, not kazu

PR #66, opened 2024-09-19
Updated reminder text, added living metal

ek488:
Updated reminder texts to their current Oracle reminder texts (many changes from "enters the battlefield" to "enters", various other reminder text updates)

Added the mechanic "living metal" (from the Transformers set)

Changed the mechanic "stare down" from expert to action

Fixed typo on "fathomless descent"


CajunAvenger:
wow that's a lot more random snags than i expected. thanks for the fixes~

PR #67, opened 2024-09-20
Symbol font module

G-e-n-e-v-e-n-s-i-S:
What slows MSE down the most are by far the blending operations. (Not because they are unoptimized, but just because they are expensive.) So I believe that for mana symbols at least, we should pre-blend everything. Now that does lead to about 5000 trybrid symbol files (with all 10 colors, snow, colorless, artifact, large and small variants). I think it's worth it, but I can be convinced otherwise.

One good option to reduce this number is to define a canonical ordering for the symbols. Right now I've rendered all combinations, so there's `W/U/B` but also `W/B/U`, `B/W/U`, etc... That would divide the number by 6 if my math is correct.

Also, I rotated the trybrids 180 degrees so they look more like the hybrids:
rotated trybrid

The chaos symbol had two bindings: `chaos` and `A`. I removed `A` to use it for artifact mana. This breaks compatibility.

For hybrid generic mana, I did `2/[WUBRG]` and `3/[WUBRG]`, but I didn't implement the `[0-9]/[WUBRG]` cause I'm lazy, and no-one uses these.

I have yet to do a pass on the existing symbol fonts, to delete redundant files and implement the module. This may be better to do in another PR.


CajunAvenger:
i am very skeptical of manual blending mana symbols, we don't really see lag time from rendering symbols, is going to be a hurdle for any new symbol font, and is probably going to remove if not completely overshadow the file size gain we get from condensing into a module

CajunAvenger:
i'm not sure i'm even going to be able to review this its so much to fetch x.x

G-e-n-e-v-e-n-s-i-S:
None of your arguments convince me, but if you feel strongly about it, I'll change it.

CajunAvenger:
are you getting any notable performance increases?

my first place to look for optimizations here would have been reducing the file size, most mana symbols are way bigger than they need to be. you cut them down to like 40x40, you don't lose any size on a 750w frame but you have about 5% of the pixels to run through the blend script

G-e-n-e-v-e-n-s-i-S:
This is reviewable. Once it's merged I can start deleting redundant files in existing symbol fonts. I think it's best if I do this in another PR.

G-e-n-e-v-e-n-s-i-S:
Yeah this branch is lost...

How are we always getting the craziest bugs from github.... first the infinite number of files changed, now this...

PR #68, opened 2024-09-21
add popouts and foils modules

CajunAvenger:
using this in #18 but wanted to pop this into its own PR to make sure @G-e-n-e-v-e-n-s-i-S gets eyes on it.

main thing i'm on the fence about is if foil_mask_blended_image() should have the default check for pt and stamps or if that should always be manually overwritten


G-e-n-e-v-e-n-s-i-S:
The `popout_coord` function should return `to_number_lax(split[input])`

Hagan said he could maybe look at my PRs next week, which means we might have the `file_exists` function ready for this update. If that happens, then checking for default masks becomes free.

CajunAvenger:
oh yep, accidentally deleted that line before the first commit and put it back incomplete in the second

hmmmm guess i'll throw another week at random cleanups. hopefully hagan can figure out what the auto update is looking for too

G-e-n-e-v-e-n-s-i-S:
There's always the corners update that needs doing, if you don't have anything on your plate.

G-e-n-e-v-e-n-s-i-S:
also, once the symbol font module is merged, we can delete redundant files in all existing symbol font

also, we should probably add options to the export templates to export all cards at 150DPI

CajunAvenger:
> There's always the corners update that needs doing, if you don't have anything on your plate.

what's the corners update?

G-e-n-e-v-e-n-s-i-S:
> what's the corners update?
currently, some frames like `magic-m15-flip.mse-style` have corners built into the card png, which means that if we want to remove them, we actually need to have the `card.corner` field display black. but that doesn't work on templates that have art there.

so the corners update consists of going over every template and removing the built in corners, replacing them with a black border. that way we can merge the "none" and "invisible" choices on the `card.corner` field, and have a single choice that removes corners on all frames.

CajunAvenger:
most of those can be handled by removing the radius on the border color rather than editing the frames, most if not all of which were done in module rollout

PR #76, opened 2024-09-24
Indicator update

CajunAvenger:
tighten up the code of has_identity so it only shows the indicator when it needs to. this almost entirely removes the need for the "color indicator dot" styling options. the two potential use cases to consider is people that want to force the cidot on for some reason, and for people that clear out their indicator manually and then need to change it later (since it won't be possible to see it anymore). 

todo is to remove the styling options from templates

includes example .mse-set file of many different color iterations interacting with the script


CajunAvenger:
do you want your suite updated or have them left in @G-e-n-e-v-e-n-s-i-S ?

G-e-n-e-v-e-n-s-i-S:
sure you can update them. I'm a bit busy today and tomorrow

PR #77, opened 2024-09-26
improve script caching

CajunAvenger:
it turns out card_number was running 7 times for every card in the set whenever a card name is changed, causing rough lag on bigger sets that were sorting the entire set 2100 times in between each key stroke. ideally we find a way to better cache this because its still sorting it 300 times but 7x faster is still good.

gonna do a run to see if we can cache anything else for additional minor improvements


CajunAvenger:
@G-e-n-e-v-e-n-s-i-S `detect_custom_creature_subtypes_statistic` and `global_variable_average_opening_hand_statistic` cause noticeable slowdown in the type field of larger sets. i've added a concept for unloading those here:

for detect_creature it just removes the automatic scanning and leaves the ability to actively scan for them by setting the field to "detect", removing the passive scan

for global_variable this is more a step 1. The stats page loading for this does seem very slow, but just making it a Set Field where you choose to load it like this is pretty quick. maybe the play is to reformat it to be nicer to read in this field rather than the Stats tab?

G-e-n-e-v-e-n-s-i-S:
totally on board for removing the auto scan on detect_creature.

more hesitant on the global_variable, but that's because it's touching the stats page my baby

CajunAvenger:
we're able to stomach one of these, so i'll leave it for now. it might be on the chopping block in future though

PR #89, opened 2024-10-02
Saga module

G-e-n-e-v-e-n-s-i-S:
- add saga module
- add classicshifted saga (1200DPI, will scale down the assets to 300DPI just before merge so it doesn't take gigabytes)
- add universes beyond saga


G-e-n-e-v-e-n-s-i-S:
The chapter icons seem to have been duplicated somhow

CajunAvenger:
the chapter numbers from your module had numbers that didn't work with replacement chapter location in Sagas and didn't get cleaned up in the quick pull of that pr

CajunAvenger:
incomplete proof of concept

1) isn't currently loading on startup because it uses text layout data. there's gotta be some way for that to work properly because flavor bar doesn't freak out but i'm blanking rn. typing a few characters will get it to sort itself out
2) doesn't update on styling.snap_chapter_to_box rn, think its weirdness on how those scripts aren't dependent on it anymore but trying to hook it back in isn't working for me
3) lower chapters are off some because i'm not calcing height quite right i'll look at that when its not midnight

PR #93, opened 2024-10-04
Add color_background mask parameters

G-e-n-e-v-e-n-s-i-S:
Added the ability to specify full black or white masks, using the `multicolor_blend`, `hybrid_blend`, and `artifact_blend` parameters.

I purposefully didn't do too much refactoring so it's easier to review (hopefully).

The idea is that whenever there is a call to the `masked_blend` function, I preface it with a check to the appropriate `multicolor_blend`, `hybrid_blend`, or `artifact_blend` parameter, to see if we can skip the call.


G-e-n-e-v-e-n-s-i-S:
I would really like it if the default value for these parameters was "white" instead of "mask", but that would require modifying all the frames.

PR #97, opened 2024-10-17
Revert watermark alpha

G-e-n-e-v-e-n-s-i-S:
Why was this changed?


CajunAvenger:
Watermark blending was at 75% before the watermark changes. This was changed because it came up as a bug for several people that their watermarks were no longer visible; people who need 25% can make the watermarks fainter as they didn't already exist.

was something built-in using 25% cause that's a crazy alpha to apply

G-e-n-e-v-e-n-s-i-S:
mmm, on the showcase frames I'm getting this with 75% alpha:

Untitled

which makes the text unreadable

PR #101, opened 2024-11-09
Suggested fix for cright field error in Modern Doublefaced Planeswalker template

mahx94:
extra card field: `cright line` made a call to `get_card_number_count(tag:card_number_tags().0)`
Console complaied that no such method exists

I assume it was an artifact from the old version of the template, and that it was intended to be changed in an earlier update.

Also, card_number_old_1() t causes a visual bug where a whitespace is missing between copyright and card number. Should i look into that?

Read code comments for more details


CajunAvenger:
@ your comment, you can edit the game files for pull requests, if it causes problems the PR is here so we can catch it

PR #109, opened 2024-12-31
Added Rules Tip Card

G-e-n-e-v-e-n-s-i-S:


CajunAvenger:
should this have a rarity symbol on the right side rather than a casting cost?

G-e-n-e-v-e-n-s-i-S:
There should be a styling option to add the rarity symbol.

CajunAvenger:
oops

PR #118, opened 2025-02-02
Add UNF full art basics

G-e-n-e-v-e-n-s-i-S:


CajunAvenger:
...just noticed the name defaults on this and the others. implementation on this looks completely backwards

CajunAvenger:
draft here
* No longer automates names, this was buggy and its something we usually don't care about automating anyway
* added Auto for land type checks, which checks the name for the Basics (potentially should add checks for Snow-Covered), then the typeline
* cleaned up the card_color hack
* fix the default type doing doubled and wrong-order types, add in a little extra so the default doesn't bootstrap itself if the card name is changed

PR #119, opened 2025-02-03
Refactor multi/hybrid/artifact blends

G-e-n-e-v-e-n-s-i-S:
Unclear if better.


CajunAvenger:
tweaked this so the logic is clearer and added a bunch of comments

PR #124, opened 2025-02-14
Fix printing bug

G-e-n-e-v-e-n-s-i-S:
Fix bug where cards where all printed on top of each other.
Increase version to 2.5.7.
noidea


CajunAvenger:
the issue was something in the other libraries having mismatched versions between us and hagan, have you tracked that down or changed code?

G-e-n-e-v-e-n-s-i-S:
I have changed code.

The changes can be seen in this PR:
https://github.com/haganbmj/MagicSetEditor2/pull/102/files

Basically, cards are first rendered to a canvas (called a DC, or Device Context), then the canvas is placed on the page that is to be printed.

The old code would move the canvas' origin to where the card needed to be on the page, then render the card at the canvas origin. I changed it so that the canvas origin always stays at the page origin, but then I draw the card at the correct coordinates inside the canvas. In my mind, this is a different way of accomplishing the exact same thing, but for some reason, it fixes the bug. (Also I had to scale the coordinates by the template DPI, for some reason.)

So yeah, as you said, I guess something in the underlying implementation of the `DC.SetDeviceOrigin`, and the `DC.DrawBitmap` functions has changed, but that's in the wxWidgets library, so I have no idea what it is.

PR #126, opened 2025-03-02
Bump to 2 6 0

G-e-n-e-v-e-n-s-i-S:
I have massive respect for the work you put into Mainframe DFC, but seeing how trigger happy Wizards has become with new frame layouts, I believe maintaining such a template is unsustainable. You certainly have the technical capacity, but I see the toll it's taking on your motivation. I understand this template isn't going anywhere, it's too ingrained in the ecosystem, however I want to start pushing for an alternative.

So I prototyped an idea: card linking.

Each card is assigned a unique id (just a string of 32 random digits), and they can now use this id to refer to up to four other cards. This creates links between them that can encode relations such as Front Face//Back Face, Generator//Token, Meld Component//Meld Result, etc...

Links can be created from the Cards menu:
Links

The possible relations are defined at the game level using the following syntax:
```
card link:
	Selected card relation // Linked card(s) relation
```
Custom relations can also be defined by the User.

These links and relations are then displayed in the GUI:
Links GUI

Right now, when you click on one of those linked cards' display, it selects the card. Maybe in the future I'll look into allowing to edit the card directly from there.

The links can be accessed inside scripts using the following syntax:
```
card.uid
card.linked_card_1
card.linked_card_2
card.linked_card_3
card.linked_card_4
card.linked_relation_1
card.linked_relation_2
card.linked_relation_3
card.linked_relation_4
```

Changes to the exe's code can be seen in this PR:
https://github.com/G-e-n-e-v-e-n-s-i-S/MagicSetEditor2/pull/30/files

Lots of bug squashing needs to happen, but first I want to know if you think this is going in the right direction, and if so, how can I improve it.


CajunAvenger:
goodness.

being able to forward edit the other face is probably the biggest step to make it replace mainframe dfc.
would like to see some way to copy the card image as a dfc like mainframe dfc does. exporters can presumably do this with card.link_card

how does the UI handle linking horizontal cards?
how does it handle copying cards into other sets?

CajunAvenger:
maybe this recontextualizes 'mainframe rebuild' as a new template. Planeswalkers can now get typeline automated if we wanted to, this can check card.linked_relation to see if it should toggle on the transform dongles.

would a backface know it's linked to a front face?

G-e-n-e-v-e-n-s-i-S:
> being able to forward edit the other face is probably the biggest step to make it replace mainframe dfc.

alright, I'll make it happen. when there is only one linked card, its preview is bigger. I'll make it so that in this case, you can edit the linked card directly. when there are multiple linked cards however, their preview is small (I could make it bigger but there wouldn't be any room left for the card list). in that case, they are too small to edit directly, so I'll keep the behavior of just switching the selection.

> would a backface know it's linked to a front face?

yes. among the backface's linked cards, there will be one labeled "Front Face".

> would like to see some way to copy the card image as a dfc like mainframe dfc does. exporters can presumably do this with card.link_card

yeah, exporters can find and render the two sides with Hagan's `to_card_image` function. I might need to add a script function to stitch two images together? as for the copy/paste, that seems more complicated, but maybe I can add another keyboard shortcut, like say ctrl+shift+c, which would copy the card and all linked cards into a single image? 

> Planeswalkers can now get typeline automated if we wanted to

I'm really not a fan of cramming too much stuff into a single template. it leads to maintainability nightmares.

> how does the UI handle linking horizontal cards?

not particularly gracefully. test it and give feedback.

> how does it handle copying cards into other sets?

a link between two cards is nothing more than them having each other's uid in one of the `card.linked_card_X` fields.
when you paste cards into a set, from the same set or another, it checks for uid conflicts, and generates new uids for the pasted cards if necessary. generating new uids may break links, but if two linked cards are pasted at the same time, and one needs a new uid, the program will update the link on the other card.
(two cards can never have the same uid, even if they are otherwise completely identical in name, mana cost, etc...)

G-e-n-e-v-e-n-s-i-S:
> being able to forward edit the other face

ok I think I mostly got this working, need to fiddle a bit to get it to refresh the card list properly

CajunAvenger:
> I might need to add a script function to stitch two images together?
that'd be a pretty big get outside of this as well

> but maybe I can add another keyboard shortcut, like say ctrl+shift+c, which would copy the card and all linked cards into a single image?
yea if it needs tucked behind a keyword shortcut or a menu option that's probably fine. potentially a button under the card in the ui?

> but if two linked cards are pasted at the same time, and one needs a new uid, the program will update the link on the other card.
very nice. needing to move both faces is a lil annoying but this is a niche enough use case that it's probably fine

G-e-n-e-v-e-n-s-i-S:
> I might need to add a script function to stitch two images together?

pretty sure we can already do this with a frankenstein of `masked_blend` and `crop` with negative numbers to actually enlarge the images, but it's probably best if I code a nice streamlined c++ version.

> needing to move both faces is a lil annoying

yeah but that only happens when pasting from one set into itself. if you paste from another set, there overwhelmingly likely won't be uid conflicts, so no need to generate new uids, and so links will be preserved even if you paste cards one at a time.

> would like to see some way to copy the card image as a dfc like mainframe dfc does.

keyboard shortcut seems easily doable actually. but it'll have to wait for later this week, got some life to attend to

G-e-n-e-v-e-n-s-i-S:
okay, I've gotten this to an acceptable state. still needs some testing, would appreciate if you could help out.

> being able to forward edit the other face is probably the biggest step to make it replace mainframe dfc

this should be fully operational.

> would like to see some way to copy the card image as a dfc like mainframe dfc does

I have made it so that if 5 or less cards are copied to the clipboard, they can be pasted into discord or paint etc as a single image file. the card images inside will be scaled down to the lowest dpi among them.

separately, I have added a keyboard shortcut (ctrl + shift + c) which will copy all cards currently selected in the card list, as well as all their linked cards, selected or not, to the clipboard.

together these two features mimic the behavior of copy pasting a dfc template.

> I might need to add a script function to stitch two images together

I have added the `insert_image` script function. here are the docs for it:
[insert_image.txt](https://github.com/user-attachments/files/19151102/insert_image.txt)

> mse 2.whatever wishlist, fixing the hang on loading unknown templates

the program should now ask for a replacement template for each individual missing template, just like it does if the set default template is missing

G-e-n-e-v-e-n-s-i-S:
TODO:
- Add built-in script functions to fetch linked cards for a given card
- Add locale entries in other languages
- testing

CajunAvenger:
head spinning at how much you've gotten here

i've been horribly sick this week so haven't been able to mess with this yet, trying to get into this asap

G-e-n-e-v-e-n-s-i-S:
no worries, get well first, nothing urgent

G-e-n-e-v-e-n-s-i-S:
I figured out how to make the `include file` directive depend on the locale.
now, if you use `include localized file:`, it will add "_" followed by the locale name to the end of the filename.
so for example, if you write `include localized file: keywords`, and the selected locale folder is `de.mse-locale`, it will try to include the file `keywords_de`

G-e-n-e-v-e-n-s-i-S:
I'm not too familiar with the various exporters. Which are the ones who would need to be modified to check for DFC card links?

CajunAvenger:
> this can wait for a later update, unless you want to do some of it here.
i will get a mainframe DFC and the relevant exporters done for this, yes

> Dark mode
hype

>  then someone else can drag that PNG into their MSE and it automagically adds all the cards with all their fields populated correctly
you can *do* that??

> Image import window remembers the last crop coordinates, and has an option to show a grid on the preview
sweet

G-e-n-e-v-e-n-s-i-S:
There is one thing I am unsure of where to put in the docs, so I will explain it here for now.

We can now paste more things in the card list:
- Text
- URLs
- PNGs
- Files

If we paste text, it will look inside this text for the following tag `<mse-card-data> </mse-card-data>`
and try to parse what is inside as a JSON array of mse card objects.

The format of a mse card object can be seen by calling the `to_json` script function on a card in the console.
It's just a JSON-ified version of what appears in `set` files, with the two differences being that
1) card fields are stored in a sub-object named `data`, rather than being members of the card object directly
2) there is an additional member named `mse_object_type` whose value is `card`.

Once it finds such text, it re-creates the cards and adds them to the set.

If we paste URLs, it will download what is at the address, and paste it in the card list.

If we paste a PNG, it will look for an iTXt meta data chunk named `Description` and paste
its contents in the card list.

If we paste a file, it will paste its contents as text in the card list.

These chain so you could paste a file whose contents are a URL at the end of which is an image
whose description is card data.

There is an additional step when pasting PNGs, whereby after creating the card, it will scan all the image field
values for a string of the form: `---X-Y-WIDTH-HEIGHT-ANGLE---`, where X, Y, WIDTH, HEIGHT and ANGLE are
integers, and interpret those as coordinates for a crop of the PNG. It will then save this crop as an image in
the set, and the image field's value will become that image's filename.

CajunAvenger:
we can have it be unstable beta just as the fork; people with github desktop installs can easily switch back and forth, without having a download link lying around for random people to find

G-e-n-e-v-e-n-s-i-S:
I dont think so. Changes here are tied to the new exe, they wouldn't work without it.

PR #133, opened 2025-04-07
Reorganise font files

G-e-n-e-v-e-n-s-i-S:


G-e-n-e-v-e-n-s-i-S:
The following fonts seem to be unused:
  Geisha Holiday W00 Regular
  vanilla whale

The following fonts seem to have the same name as far as MSE is concerned:
  MagicMedieval
  Scrolls

G-e-n-e-v-e-n-s-i-S:
After a bit of benchmarking, I find no noticeable difference in boot time between loading all the fonts including subfolders vs loading only the main folder, so we could definitely make it load everything

PR #138, opened 2025-05-14
Leveler module

G-e-n-e-v-e-n-s-i-S:


CajunAvenger:
are swap_fonts_level2/3 actually used? Looks like this module isn't supporting DFC frames since we'll be downsizing those soon, so don't think these need to be here.

kind of want to remove the oddball _1 from the script names too

(things i'll probably get to first)
* Custom level font should prob be part of leveler module rather than fonts? or be a secondary include file in there
* the default pt font size is 300 which seems incorrect
* Colorless on UB Levelers has a miscolored rectangle around the stamp
* If we're gonna make the levelboxes selectable anyway i wanna try to get loyalty costs and ptbox options in here, particularly cause this'll make it harder to do that for other templates to add it otherwise. We're also missing a style option to change the default
* I also wanna get the "lone pt leveler" behavior in here
* As with walkers, investigating an auto-levels option, Spacecraft have made that more relevant

CajunAvenger:
> will you provide an external way to modify sets at https://mse-editor.neocities.org/ to break down cards using Mainframe DFC into their constituent parts?

yes. i'm planing to leave the frame, at least for now, and handle mse-editor splitting that into two mainframe cards

> my plan is to remove them before merge.

o, k.

> the font size is actually determined by the field height. 

at least for my computer, the cursor size isn't, so i'd have a 300px blinking cursor for a 16px text

> also it kind of shifts the text upward because of the last line with the zero width space.

you don't zero-width the last field. this lets you get it with down-arrow without it moving the textbox, like how it ft works for cards with a ton of rules text

G-e-n-e-v-e-n-s-i-S:
> i'd have a 300px blinking cursor for a 16px text

weird, that doesn't happen on my end. well definitely needs changing then.

> you don't zero-width the last field

this is the behavior I'm getting with mainframe walkers. on the left is "three abilities", on the right is "automatic":
walker_text_shift

CajunAvenger:
its much less noticeable in practice where you have 6 or 9 lines of text. plus its keying off a style option so i could just make the textbox 3px taller in that case to compensate if we really care :v

PR #141, opened 2025-05-15
Bump to 2 5 8

G-e-n-e-v-e-n-s-i-S:
- Hard code exception for word "plains" in `english_singular` and `english_plural` functions
- Add font preloading by TomTkacz
- Add `import_image` function by TomTkacz
- Fix freeze when loading sets with missing templates
- Add `include localized file` token
- Add locale keys

### DIFF
--- a/data/ch-s.mse-locale/locale
+++ b/data/ch-s.mse-locale/locale
@@ -52,6 +52,10 @@ menu:
 	add card:							添加卡牌	Ctrl+Enter
 	add cards:							批量添加卡牌...
 	remove card:						删除所选卡牌
+	#TODO: Localize
+	link card:							&Link Cards To Selected Card...
+	#TODO: Localize
+	copy card and links:				Copy Selected Cards And Their Linked Cards	Ctrl+Shift+C
 	orientation:						旋转
 	rotate 0:							普通
 	rotate 270:							顺时针旋转90°
@@ -171,6 +175,10 @@ help:
 	add card:							添加一个新的空白卡牌到本套牌
 	add cards:							添加多张卡牌到本套牌
 	remove card:						从本套牌中删除所选卡牌
+	#TODO: Localize
+	link card:							Link one or more cards to the selected card
+	#TODO: Localize
+	copy card and links:				Copy the selected cards and all their linked cards
 	orientation:						调整所显示卡牌的方向
 	rotate card:						将所显示卡牌顺时针旋转90°
 	rotate 0:							将所显示卡牌还原为正常模式
@@ -391,6 +399,10 @@ tooltip:
 
 	add card:							添加卡牌
 	remove card:						移除选中卡牌
+	#TODO: Localize
+	link card:							Link cards to selected card
+	#TODO: Localize
+	copy card and links:				Copy selected cards and linked cards
 	rotate card:						旋转卡牌
 
 	add keyword:						添加关键词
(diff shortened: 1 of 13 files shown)


CajunAvenger:
could we get `include localized file:` to fall back to the normal file name and proc a warning if the localized one doesn't exist?

G-e-n-e-v-e-n-s-i-S:
I wanted to do that, but it requires digging into the package management code, which is a complete nightmare, so I'm just not gonna bother.

CajunAvenger:
dang. might wait on utilising that in the pack til next major version then. still good to have

G-e-n-e-v-e-n-s-i-S:
for keywords and statistics, it's just a manner of renaming the files.
here's a commit that does that, feels free to revert it:

CajunAvenger:
> for keywords and statistics, it's just a manner of renaming the files and creating da and pl versions (just copy the en one).

it's not about creating those files, but the potential of hard breaking things if someone gets a partial update or has a custom locale

> Also, I was thinking maybe we could move the set.card_language field to the app level, just next to the app language. Add a script function to fetch it.

i guess? maybe have a null/default value so older files can fall back to the locale (in the exe) and set.card_language (in the scripts) rather than make everyone re-set it


does import_file have a way to import local files without needing the whole filepath? wondering if we could replace the "can't find image, try saving and reloading" with an attempt to import that image

G-e-n-e-v-e-n-s-i-S:
> it's not about creating those files, but the potential of hard breaking things if someone gets a partial update

if they get a partial update they will have so many errors that adding one to the pile won't change anything. plus this particular error will lock them out of the program, preventing them from damaging their set files.

> or has a custom locale

that seems like an extremely niche problem

> i guess? maybe have a null/default value so older files can fall back to the locale (in the exe) and set.card_language (in the scripts) rather than make everyone re-set it

yeah this is probably more trouble than it's worth, let's just shelve this suggestion

> does import_file have a way to import local files without needing the whole filepath?

I can add a `get_mse_path` script function that returns the absolute path of the app folder

G-e-n-e-v-e-n-s-i-S:
Small update to the exe:
 - Add `get_mse_path` script function which returns the absolute path of the mse app folder
 - Stop dropdown menus from closing when there is a console message
 - Dropdown menus who's choices are numbers are now displayed as sliders
 - Added `slider` field type, which is just syntactic sugar for a choice field, so the following two fields are essentially identical:
 ```
 card field:
	type:			slider
	name:			test
	minimum:			12
	maximum:			18
	increment:			2
card field:
	type:			choice
	name:			test
	choice:			12
	choice:			14
	choice:			16
	choice:			18
```

G-e-n-e-v-e-n-s-i-S:
- Added CSV / TSV importer, available from the `Cards` menu.
- Added `alt name` field property. The CSV importer will look for these names when parsing, to know which field to populate.
- Added `construction script` field property. This script will be applied to the value found in the CSV. It must return a map of field names to values. For example, the `pt` field should not be populated directly, since it's just a mirror of the `power` field and the `toughness` field. So if a value is supplied for it in the CSV, it should be broken down and redirected to the underlying fields, like so: `{split := split_text(value, match:"/"); [power:split[0], toughness:split[1]]}`. (This script actually applies whenever the `new_card` function is called, so it also works in the CLI and the `add_cards_script`, although it is likely not needed there, since we can control exactly which fields are populated.)
- Added `construction script` game property. This script is run as the final step of the `new_card` function, to do some global cleanup. It must also return a map of field names to values.
- Added `stylesheet` and `notes` keys to the `new_card` function, so that these values can be populated.
- Added `make_map` script function, which takes a list of keys and a list of values, and produces a map. This allows for dynamical map initialization.
- Added some basic `alt names` and `construction scripts`, which we can expand upon later.
- Added locale entries (localized french, italian, german and portuguese)

G-e-n-e-v-e-n-s-i-S:
and with that I'm gonna stop piling shit here god dammit I always do this and end up with thousand line PRs you must hate me so much lol

CajunAvenger:
> am I crazy or was there no way to do this from the CLI?

does new_card not work from the CLI or does this do something else

G-e-n-e-v-e-n-s-i-S:
> does new_card not work from the CLI or does this do something else

it creates a new card but does not add it to the set

PR #142, opened 2025-05-16
Update french keywords

G-e-n-e-v-e-n-s-i-S:


CajunAvenger:
is un/une common enough to get a parameter? instinctively i want gendered params to deal with this but i'm not familiar enough with the relevant languages to know if that's even reasonable

PR #152, opened 2025-08-14
Repair and partially update old export to html option.

GoldenFlame0:
The previous exporter was nonfunctional; I spent a little bit of time cleaning it up, adding missing functionality, and updating the CSS.


GoldenFlame0:
Split cards don't really work well at the moment (especially rooms) and I haven't tested double faced cards, but this should be at least a start.

GoldenFlame0:
Thankyou Cajun!

Hopefully I've made enough of a headstart here for anyone else who's interested to finish (I got stuck on some of the field names), but at least now this old dinosaur is back where it belongs.

PR #182, opened 2026-03-04
[develop] integrate dark mode in templates

G-e-n-e-v-e-n-s-i-S:
integrate dark watermarks in templates

### DIFF
--- a/data/magic-DayNight.mse-style/style
+++ b/data/magic-DayNight.mse-style/style
@@ -305,7 +305,7 @@ card style:
 		z index: 1
 		render style: image
 		alignment: middle center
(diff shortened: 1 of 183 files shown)


CajunAvenger:
the cache is having some trouble when switching between dark mode and light mode

whichever version you first load up is used for both light mode and dark mode. it looks like the cache might be trying to save both, but the second image is corrupted (or maybe that's the none icon being weird)
image
image

G-e-n-e-v-e-n-s-i-S:
that does indeed look like uninitialized data which is terrifying.

does it automatically go away after you use the same mode for a while?

CajunAvenger (approved):

PR #185, opened 2026-03-04
[develop] merge archenemy, vanguard, planechase, herospath into magic

G-e-n-e-v-e-n-s-i-S:


CajunAvenger:
should the new card fields just be extra card fields

ed: actually archenemy/planechase frame should be Style options; that's something you want to set once and not change every time

G-e-n-e-v-e-n-s-i-S:
I agree they should be styling fields but there is a problem that made me switch to card fields, which is the `vanguard_hand_mod_label_default()` and `vanguard_life_mod_label_default()` functions don't get re-run when you change the frame if it's a styling field. They only get re-run when you reload the data.

(then I switched the other three games to the same system for consistency)

Do you know a workaround for this?

CajunAvenger:
planechase has some rounding error issue creating this white line in the bottom right in MSE but not export renders
image

ed: i don't love that "Original" isn't actually a version of the frame that was ever used. you can get the actual Original and the new MOM version with the other style options but that should probably be a baseline thing and not flipping a bunch of options.

G-e-n-e-v-e-n-s-i-S:
> planechase has some rounding error issue creating this white line in the bottom right in MSE but not export renders

no sense fighting against this, it depends on the zoom% at which the card is displayed. there will always be zoom values where that happens.

as long as the card is flawless when exported at native resolution, we're good.

CajunAvenger (approved):

CajunAvenger:
i'll do the update cards scripts for mainframe planes and m15 vanguard in a follow up PR, emblems already has a utility function for it that i'll pull when that's on develop. I'll get mse-editor to handle moving the alternate games to the magic versions

PR #187, opened 2026-03-05
[develop] add japan and through the ages showcase frames

G-e-n-e-v-e-n-s-i-S:
also revamped the playtest frame
also fixed a bug in the ixalan full art frame, and another in the the tarkir dragon wing frame


CajunAvenger:
not a fan of the name "Art Upload", i thought this was a copy of the Secret Lair Full Art frame at first. Through the Ages is kinda unhelpful too tho, maybe "Source Material" like scryfall uses? its preview image should prob be an example card too rather than only the bottom border.

G-e-n-e-v-e-n-s-i-S:
source material does sound much better. I'll let you find a preview image

CajunAvenger (approved):

PR #188, opened 2026-03-05
[develop] refactor bullet point spacing

G-e-n-e-v-e-n-s-i-S:
looks like there's a bug in the exe that I need to address before this can work correctly

### DIFF
--- a/data/magic-crystal-cave.mse-style/style
+++ b/data/magic-crystal-cave.mse-style/style
@@ -252,10 +252,10 @@ card style:
 		right:				{ 666 - chop_right() }
 		bottom:				{ 940 - chop_bottom() }
 		z index:			900
-		line height hard:	{ if is_modal(card.rule_text) then 0.9 else 1.2 }
+		line height hard:	1.2
 		line height line:	1.5
 		line height soft:	0.9
-		line height hard max:	{ if is_modal(card.rule_text) then 1.0 else 1.3 }
+		line height hard max:	1.3
 		line height line max:	1.6
 		alignment:
 			script:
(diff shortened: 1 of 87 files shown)


CajunAvenger:
we should probably auto-add the <soft-line>s into modals? keep it working like live and idk that we need bigger-than-hard-linebreak-modal-breaks

G-e-n-e-v-e-n-s-i-S:
I don't understand what you mean but sure go ahead

CajunAvenger:
currently hard line breaks inside a bulleted list are forced into soft-line tags (i think github ate the code tag before). this version isn't, so if you don't manually soft-break you get the 1.2 line height plus the additional margin, instead of 0.9 line height line you get on live, or the 0.9 plus margin that this PR was partially made to achieve.

G-e-n-e-v-e-n-s-i-S:
I see, yeah probably force soft breaks then, but this has to be done from the game script, not exe. Other games might want to be able to do this distinction.

CajunAvenger:
ye, just wanted to clear that before pushing changes

CajunAvenger:
k this should be good whenever you're ready to merge it

CajunAvenger:
> this file will get deleted by another pr so don't change it or it will cause conflicts

we'll be reinstating at least a shell of them for the update cards scripts won't we

G-e-n-e-v-e-n-s-i-S:
if you want to do that work then fine, but I'm not sure it's worth it for seldom used templates like these

CajunAvenger:
i have about 60 set files using that template I'll def spend a few minutes on an update script over the time to manual convert all of those lmao

CajunAvenger (approved):

PR #192, opened 2026-03-22
[develop] transformation module

G-e-n-e-v-e-n-s-i-S:


G-e-n-e-v-e-n-s-i-S:
Here is my idea for the transform module.

I have broken down the symbols into three parts: background, rim and icon:
3parts

that way each part can be customized separately by the user. This allows for example to make the colored versions that the "sparker" had, but for all the symbols.

It also allows me to make variants for the showcase frames rather easily. I have implemented it here for the avatar elemental frame, the ixalan caverns frame, and the bloomburrow woodland frame. All I need to do is create one image for the background, and one image for the texture of the rim and symbol:
assets

possibly with color variants if needed. The color variants are fed through the blending code so they blend with the card color that is behind them.

I probably as always made things way more complicated than they need to.

CajunAvenger:
Right-handed transformation symbol mirrors the symbol icon, which i think is incorrect. we don't technically have any canon examples that aren't symmetric already but imo it shouldn't do that.

i kinda wish the require_exclusive_choice list in transform_symbol_script was cleaner but there's not really a "one source of truth" version we can do, and we're proobably not gonna need to make the list longer. (also is choices a new parameter, that doesn't seem to be on the docs)

looks like this has custom symbols just using the old system, i'm gonna take a stab at letting them be used in the new thing, either a style option or exists_in_package check to see if the icon/shape folder is there

CajunAvenger:
the GIMP script just fails without any acknowledgement if it gets an image that's not the right size, if that's something that's fixable in there. otherwise i'm adjusting the readme instructions to note this.

i'm setting the mainframe-extras folder as /shape instead of /icon/shape since the plugin only makes /shape

G-e-n-e-v-e-n-s-i-S:
sure, I'll make it skip images that are not of the correct dimensions and issue a warning

CajunAvenger:
shuffled is_mirrored to is_right, added is_mirrored as a way to flip the icon if you wanna

currently no way to do that in the field, just a script that the style can edit, idk if you wanna put that in the options or just leave it like that

G-e-n-e-v-e-n-s-i-S:
So is this acceptable? Can I merge it into develop and start implementing it into the showcase frames?

CajunAvenger:
yup, good whenever you're ready

PR #200, opened 2026-04-23
[develop] Refactor emblem frames

G-e-n-e-v-e-n-s-i-S:
Combine all emblem frames into one, except for the oldest one.


CajunAvenger:
two more wibbles in update cards script

1) new_card doesn't accept images, so you can't currently do things like set card.image_2 as a new card's card.image. import_image("internal file name") doesn't seem to work, and we don't have a way to grab the file name anyway. possibly the line is letting import_image take an image in place of a path?

2) setting styling_data of type "package choice" prompts the error: `Could not set current working directory (error 161: The specified path is invalid.)`

(also clear for rebasing)

G-e-n-e-v-e-n-s-i-S:
pushed a change on develop to address these two problems. you can safely rebase.

import_image() should now accept images created with scripts, so you can do stuff like:
`new_card([image_2: import_image(saturate(card.image, amount:1))])`

as for the package choice problem, it seems to go away if you don't put a leading "/". I wrote some code to strip it automatically, but this may be a hack and the problem may resurface later.

G-e-n-e-v-e-n-s-i-S:
ok the package choice fix hadn't actually gone through cause I'm a potato but it should be good now

CajunAvenger:
all good, i haven't had a chance to rebase yet anyway

CajunAvenger (approved):

G-e-n-e-v-e-n-s-i-S:
in the deprecated warning, we should probably use the short and full names instead of the folder name?

otherwise seems good, you can merge whenever

CajunAvenger:
maybe it should use the path needed to find it in the installer window?

G-e-n-e-v-e-n-s-i-S:
This frame has been replaced by: short_name - full_name

Which can be found in the package manager under: installer_group

Cards will switch to that frame once it's installed and the program is re-launched

PR #210, opened 2026-05-12
update cockatrice exporter

ebbit1q:


ebbit1q:
I have not tested this yet, please provide feedback on if my msescript makes sense, thanks

ebbit1q:
I noticed my pr conflicts with #209 I'll most likely have to remake my changes soon

CajunAvenger:
i don't think they actually conflict? 209 is primarily refactoring token associations to handle a new feature from 2.6.0

ebbit1q:
thanks for merging this! can someone maybe help me figure out what the "splitter path" is supposed to do? when unescaped xml characters end up in there it breaks the file but I don't want to break the script either! splitter path is not a recognised value by cockatrice

CajunAvenger:
> thanks for merging this! can someone maybe help me figure out what the "splitter path" is supposed to do? when unescaped xml characters end up in there it breaks the file but I don't want to break the script either! splitter path is not a recognised value by cockatrice

i think it used to be for an external exe that would split DFC images and update the XML. that can 100% safely get xml_escaped() and like 99.99% safely deleted, MSE is capable of doing that on its own now and the main community that was using that tool collapsed, idk if anyone is still using it.

PR #212, opened 2026-05-19
optional parameter support

CajunAvenger:
While mse theoretically had support for optional keyword parameters, the placeholder text would still be rendered, eg "token\<s>", so instead we used the inelegant match "toke<*s>". This PR instead adds a step to text_filter to remove placeholders with a standard overwritten name, drafted as `<atom-kwpph>hidden</atom-kwpph>`

This does mostly function, "Type token\<plural>" and "Connive\<plural>\<optional_number>" both work correctly. But there are two weird interactions:
* If there is any plaintext—including just a space—after the optional param, it becomes mandatory. "a\<plural>b" matches "asb" but not "ab". This prevents updating into "become\<plural> plotted".
* If there are any parameters separated from the optional by plaintext, it becomes mandatory. This prevents updating into "mill\<plural> \<one_word> card\<plural>"

this is not returning `keyword` in expand_keywords(), so i think it's a hiccup in the matching, rather than some quirk in expand_keywords that returns no reminder text, and the script step that removes the empty placeholder def isn't involved. The only time `optional` is actually [referenced](https://github.com/G-e-n-e-v-e-n-s-i-S/MagicSetEditor2/blob/2f77e8c4c1b77b898752a1bc1e76dfd1d8993100/src/data/keyword.cpp#L189) is adding a `?` to that part of the regex, so I wouldn't think it's some bug with optional placeholders. My instinct is that it's something happening in the filtering happening at [possible_matches](https://github.com/G-e-n-e-v-e-n-s-i-S/MagicSetEditor2/blob/2f77e8c4c1b77b898752a1bc1e76dfd1d8993100/src/data/keyword.cpp#L388), but i can't follow it well enough to tell where

this is probably still an acceptable update with those examples left on the old method, but wanna make sure @G-e-n-e-v-e-n-s-i-S gets a chance to check the quirks before doing that.

### DIFF
--- a/data/magic.mse-game/keywords_en
+++ b/data/magic.mse-game/keywords_en
@@ -173,6 +173,24 @@ keyword parameter type:
 	separator before is: [^>]([Pp]ays?( an additional| any amount of)?|gets?( that many)?)[ ]
 	reminder script: length(input)
 
+keyword parameter type:
+	name: optional_number
+	match: [ ]?[XYZ0-9%]+
+	optional: true
+	placeholder: hidden
+	refer script:
+		name: corrected value
+		script: \{clear_optional_placeholder({input})\}
+
+keyword parameter type:
+	name: plural
+	match: s|es|ies|en
+	optional: true
+	placeholder: hidden
+	refer script:
+		name: is plural
+		script: \{is_plural({input})\}
+
 ############################# All Magic keywords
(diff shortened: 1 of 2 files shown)


G-e-n-e-v-e-n-s-i-S:
I'm sorry this is another days long detour into figuring out how the keyword parsing works that I can't afford.

CajunAvenger:
you got it

PR #217, opened 2026-05-22
Improvements to Omen style layout for M15 Adventure Mainframe

Aresiel:
I updated the style for the m15 adventure template to better work with the omen left/right styles. I did at least:
- Vertically align the omen style textbox better, especially noticeable with smaller text.
- Make the width of the field (with the right: attribute) reasonably long, so that it doesn't overlap the casting cost or go outside the card. This was primarily for omen but applied to flat and page too.
- Moved the casting for omen on page 3 further to the right, to better match the prepared look.
- Horizontally aligned the name and type of omens.

I've tested to make sure I didn't break the page or flat while doing this, but probably good for someone else to double-check as well. I started using Magic Set Editor at around 1am this morning.

Some problems might occur if one attempts to use omen styles with spotlight, but spotlight doesn't actually support anything except for page anyways so that should be fine.

### DIFF
--- a/data/magic-m15-adventure.mse-style/style
+++ b/data/magic-m15-adventure.mse-style/style
@@ -1,4 +1,4 @@
-﻿mse version: 2.0.0
+mse version: 2.0.0
 game: magic
 short name: M15 Adventure
 full name: Mainframe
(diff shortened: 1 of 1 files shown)


Aresiel:
This has bugs with only 1 special page I did not yet find, I'll update when they're fixed.

Aresiel:
I believe I've fixed it now, tested more thoroughly too and it seems to work.

PR #231, opened 2026-07-07
Add workflow to automatically update other repos

G-e-n-e-v-e-n-s-i-S:
This PR should be merged directly into main, just before we merge the 2.6.0 update.

It adds a job that will be run automatically on every push to Full-Magic-Pack/main. The job will check which packages in the data folder have been modified by the push, rebuild their installer, update the installer list, then push those changes to the Installer-Pack repo. It will also mirror any change made to these packages to the Basic-M15-Magic-Pack and M15-Magic-Pack if they also contain the package.

(This should probably be tested with a small push before we merge 2.6.0.)

Since this is modifying repos other than the one it is in, it needs a special access permission. Here is how to set it up. It should be done before this PR is merged.

First, create a Personal Access Token:
- Sign in to GitHub with the MagicSetEditorPacks account
- Go to Settings > Developer settings > Personal access tokens > Fine-grained tokens
- Click Generate new token
- Configure it like this:
  - Name: Syncing Actions
  - Expiration: No expiration
  - Repository access: Only select repositories
		  then select: MagicSetEditorPacks/Installer-Pack
		  and: MagicSetEditorPacks/M15-Magic-Pack
		  and: MagicSetEditorPacks/Basic-M15-Magic-Pack
  - Permissions: Add permission: Contents
		  then change access to read and write
  - Click Generate token and copy it somewhere immediately — GitHub won't show it again

Second, create a Secret from the Personal Access Token:
- Go to MagicSetEditorPacks/Full-Magic-Pack
- Settings > Secrets and variables > Actions
- Click New repository secret
- Name it exactly: SYNCING_ACTIONS_TOKEN
- Paste the token as the secret
- Click Add secret

You can see on lines 25, 32, and 39 of the yml file that i grant this SYNCING_ACTIONS_TOKEN to the workflow, which means it will have read/write access to these 3 repos.

### DIFF
--- a/.github/workflows/sync-repos.yml
+++ b/.github/workflows/sync-repos.yml
@@ -373,18 +373,20 @@ jobs:
                   package_changed = any(p.replace("\\", "/").startswith(package_prefix) for p in changed)
                   if package_changed:
                       #### Update the main pack if necessary
-                      if package.name in main_repo_packages:
+                      if package.name in main_repo_packages or re.fullmatch(r"magic-.+\.mse-include", package.name) or package.name.startswith("magic-m15"):
                           t0 = time.time()
                           main_repo_package = main_repo_data / package.name
-                          shutil.rmtree(main_repo_package)
+                          if main_repo_package.exists():
+                              shutil.rmtree(main_repo_package)
                           shutil.copytree(package, main_repo_package)
                           print(f"::notice::package {package.name} synced into Main-Magic-Pack in {time.time() - t0:.1f}s.")
 
                       #### Update the basic pack if necessary
-                      if package.name in basic_repo_packages:
+                      if package.name in basic_repo_packages or re.fullmatch(r"magic-.+\.mse-include", package.name):
                           t0 = time.time()
                           basic_repo_package = basic_repo_data / package.name
-                          shutil.rmtree(basic_repo_package)
+                          if basic_repo_package.exists():
+                              shutil.rmtree(basic_repo_package)
                           shutil.copytree(package, basic_repo_package)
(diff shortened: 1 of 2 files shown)


CajunAvenger:
how does the sync handle a new Showcase frame that gets added to M15 but not Basic?

G-e-n-e-v-e-n-s-i-S:
the sync doesn't currently add anything. you have to manually add it to M15 yourself. once it's added, it will update it if it gets changed.

(the actual work done is written in python, it starts on line 53)

we could, for example, add a comment to the style files to specify in which packs a stylesheet must go, and have the sync read that and add it automatically

CajunAvenger:
is it able to read the package name? magic-m15* can safely go into m15 and *.mse-include can safely go everywhere

G-e-n-e-v-e-n-s-i-S:
done

CajunAvenger:
api token set up

CajunAvenger:
the two smaller packs have been synced to Full Pack. Once Installer Pack is ready for this to start happening, we can merge this one; ideally we don't merge other changes to main until after this one is applied.

G-e-n-e-v-e-n-s-i-S:
Yeah, the idea was that we merge this in main, then we merge everything else in develop, then we merge develop in main, and that will automatically initialize the Installer Pack, since basically every template gets changed in 2.6.0 because of this PR: https://github.com/MagicSetEditorPacks/Full-Magic-Pack/pull/202

Which means we don't need to do anything special with the Installer Pack now, it's already ready.

When I said that this doesn't add anything, that's actually only true for the M15 and Basic pack. It WILL add installers to the Installer Pack if they don't already exist.

PR #238, opened 2026-07-15
[develop] changelong and lingering TODOs

### LINKED ISSUE #236: localized numbers don't add hints
spanish_number etc should return `{singular value}<hint-1>` and `{plural value}<hint-2>` rather than just the value

i think there's also some value in adding `<hint-m>/<hint-f>` for gendered languages, though this will usually be more relevant for process_english_hints than the number scripts

CajunAvenger:
this started as adding the changelog entry for updating the sattelite games via neocities and kinda spiraled.

> • Added basic CSV / TSV / JSON importers, available from the Cards menu. Scryfall JSON files are supported. In a later update, fully customizable user written importers will be available.

is user written still a "in a later update" thing?

> • Script Functions: Added `import_image` and `get_mse_path` scripts to allow MSE to load external images. In a later update we will use these to avoid needing to reload when changing magic-mainframe-extras files.
> • Updated types and keywords to Secrets of Strixhaven //TODO get us to Super Heroes
> • Localized number scripts now insert number hints //TODO actually do that

i'll get those


G-e-n-e-v-e-n-s-i-S:
> is user written still a "in a later update" thing?

it is

CajunAvenger:
k version of the custom-image-importer i got working is a little scuffed but at least an upgrade. up to you if you want to try to tune it up or merge it as is

* problem 1: we aren't caching the filepath, just loading the image, so we have to import_image every time the image is regenerated.
* solution 1: add a new state "File loaded" so the various calls know to use import_image when needed. Set fields will re-run on startup and update to "File found"
* problem 2: Styling fields will not re-run and so would try to use import_image forever
* solution 2: add can_import parameter to the verify scripts so Styling fields act the same as before
* problem 3: import_image can't or-else into a fallback, it just error messages. this also causes the typed letter to not render in the textbox which is confusing

this PR is ready now, the final TODO will just be for the forum patch notes

G-e-n-e-v-e-n-s-i-S:
I pushed a change on develop to make import_image work with "or else"

CajunAvenger:
k, imports now fall back to a small magic.mse-game icon and abort if the image_width is too small

G-e-n-e-v-e-n-s-i-S (approved):

PR #257, opened 2026-08-18
[mainframe-transform] Split refactor

G-e-n-e-v-e-n-s-i-S:
This frame's elements were very entangled with each other. I tried re-writing it to make them more independent, so they can all be mixed and matched without restrictions. It was going great, then towards the end when I added all the little corner cases that the original frame would allow, the code got messy. Overall still a net gain I think but not as much as I would have hoped.

I renamed the "use fuse arrows" styling field to "auto frames: arrows" because that's what it was basically doing.

I merged the "room images" and "other options: spillover image" fields into a single "image layout" field.

The "short room image" and "room backfill" fields have been renamed to "typebox opacity percentage" and "typebox solid color" respectively, and I added similar fields for the nameboxes and textboxes.

I separated the room textbox and the fuse textbox so they can both be present at the same time.

I added the transform module, reminder, color multiply, as well as the vehicle and snow frames.

I did not add the transform notch because there simply is no space on the card for it (and also big lazy let's not kid ourselves.)

I also fixed a problem with the flip frame's typeline width and a few pixels on the aftermath hybrid_blend_mask.

Alright that's all you get from me unless there are problems here I'm off to try and squash a few exe bugs.


CajunAvenger:
this has gotten really hard for me to touch up so have bullet points
* position hint is at 110003 for some reason
* the room typebar default opacity should be like 70% and i am not sure how it make it actually do that. is_default() doesn't seem to work on styling fields so there doesn't seem to be a way to tell the difference between a room intentionally set to 100% and the default for non-rooms
* this frame wants the reminder_disabled := { not is_linked_dfc() } real bad
* (might be fixed on develop) since image 2 has a higher z index, clicking on the right side in one image mode clicks that one and so your image just appears to get eaten. similarly room box is clickable

G-e-n-e-v-e-n-s-i-S:
> position hint is at 110003

this is according to the new numbering from this pr: https://github.com/MagicSetEditorPacks/Full-Magic-Pack/pull/202
we can discuss it there once everything else is done

> is_default() doesn't seem to work on styling fields

can you provide steps showing is_default not working?

> this frame wants the reminder_disabled := { not is_linked_dfc() }

why so?

> (might be fixed on develop)

fixed on develop

CajunAvenger:
> > this frame wants the reminder_disabled := { not is_linked_dfc() }
> 
> why so?

because it overlaps the fuse textbox, and while the fusebox is something you'll use about 1 in 3 times you use this frame, and the modal dfc hints on a non dfc split card is a thing you'll use about 1 in a million times you use this frame

> can you provide steps showing is_default not working?

...maybe typebar in specific is haunted? brand new file, is_default always returns false for styling.typebar_opacity_percentage, but also checked styling.textbox_opacity_percentage this time and *that* one seems to be working
image

G-e-n-e-v-e-n-s-i-S:
oh, the problem is actually that the update_cards_script is ran, and it sets the value for that field, which makes it exit the default state. since the card is just created, it has no stylesheet version, which is interpreted as version 0.0.0, which qualifies for the script to run.

the new commit I just pushed fixes that by setting the field's value to nil, which means "don't do anything", which preserves the default state, but maybe we want to prevent this more systematically?

G-e-n-e-v-e-n-s-i-S:
> the modal dfc hints on a non dfc split card is a thing you'll use about 1 in a million times

well, this box has some nice design space, you can hijack it for a custom mechanic, and I put a styling field to move it around the card

CajunAvenger:
> oh, the problem is actually that the update_cards_script is ran, and it sets the value for that field...
> 
> since the card is just created, it has no stylesheet version, which is interpreted as version 0.0.0

it's doing that and breaking is_default even on a brand new file?

G-e-n-e-v-e-n-s-i-S:
yes simply by setting the field's value, it makes it exit the default state

I'll just prevent update_cards_scripts from running on new sets

PR #260, opened 2026-08-24
256 fix matrix font

### LINKED ISSUE #256: MATRIX.TTF font file poor conditioning causes alignment problem
I built MSE on Linux using [GenevensiS's repo](https://github.com/G-e-n-e-v-e-n-s-i-S/MagicSetEditor2) at commit `ba07f46856f911ea2dc5759e0ba1675e3d09ecaa`. In that build, templates using `MATRIX.TTF` bundled with this repo had text alignment problems.

**Problem:** It turned out that the reason is that the vertical offsets in the font are messed up. Specifically, the horizontal header table (hhea) contains all zeros. Because of that, the rendering engine thinks the fonts are infinitely short and aligns the bottom of the characters at the top of the text box. I do not know why this problem is absent in my (older) Windows install, but my guess is that the rendering engine in that particular version/platform of wxWidgets does something smart when the hhea table values are all zero.

**Solution:** I fixed this on my install by modifying the font file using Python's `fonttools`.

I will post a PR with the fixed font file. However, I need to know this repo's compatibility policy because changing the hhea table may cause text using that font to move vertically even in cases where it's working ok now (e.g. on Windows installs). Do I need to guarantee pixel-perfect backwards compatibility? Generally, what level of testing do I need to show for the PR?

DanielSank:
Fix hhea table ascent and descent values in MATRIX.TTF.
    
    Fixes #256
    
    The ascent and descent values in MATRIX.TTF were both zero. Apparently,
    in some versions of wxWidgets or on some platforms, the vertical
    placement of text boxes is computed using those values. We saw that text
    boxes using "Matrix" font, which is provided by MATRIX.TTF, would cut
    off the tops of text. For example, an "M" would look like this:
    
    ```
            |\   /|
            | \ / |
         ___|_____|______________________
        |   |     |                      |
        |                                |
        |                                |
        |________________________________|
    ```
    
    i.e. only the very bottom of the letter would render inside the text
    box with the rest cut off.
    
    This commit adjusts the ascent and descent values to fix the problem. My
    choice of the ascent value was made by adjusting until text appeared to
    render in the right place. The descent value was chosen arbitrarily and
    should maybe be the subject of further investigation.


G-e-n-e-v-e-n-s-i-S:
it's worth confirming that there is no duplicate of this font anywhere in the font folder, otherwise it's a coin toss as to which one MSE uses.

DanielSank:
1. I can check the other fonts to make sure the names are unique. Will do that.
2. Seeing as diff's for font files are useless, and given that I yolo'd the descent value, I'm thinking it could be worth adding a readme to the fonts directory with a bit of commentary to help future devs. Yes/no?

Also, I chose the ascent value to put the type line in the right place in a magic-new template. I did *not* go for reproducing the existing vertical alignment and I certainly didn't attempt pixel perfect backwards compatibility. Where do y'all want to hold the bar on that?

CajunAvenger:
> 1. I can check the other fonts to make sure the names are unique. Will do that.

i already did that. is everything still working on your end with those fonts removed?

> Also, I chose the ascent value to put the type line in the right place in a magic-new template. I did not go for reproducing the existing vertical alignment and I certainly didn't attempt pixel perfect backwards compatibility. Where do y'all want to hold the bar on that?

i'll double check some more frames but it doesn't seem to have affected the Windows usage of the font at all.

CajunAvenger (approved):

PR #265, opened 2026-08-28
[develop] Dependency purge

CajunAvenger:
* Updated all templates to magic.mse-game version 2026-08-28. not strictly necessary but anyone getting a new template should be getting the new .mse-game, especially since
* Updated all templates using the reminder module to version 2026-08-28, and magic.mse-game as well because of the mulitply module
* Also updated blends because of the new split card blend
* deleted all the .bats we don't need anymore
* killed watermarks-counter and watermarks-counters
* removed magic-watermarks, moving its dependents to magic-modules instead

going forward we should treat these as actual dependencies, ie update the date when, but only when, we need to. there's no need to do a big sweep like this every time we touch magic.mse-game, but also the last time that happened was 2008 we were overdue


CajunAvenger:
finished

G-e-n-e-v-e-n-s-i-S:
this looks like it's built on the wrong base. it is overwriting some headers when it shouldn't, for example in:
magic-b1234.mse-style
magic-counter-m15.mse-style
magic-m15-aftermath.mse-style
etc...

CajunAvenger:
ouagh rebase didn't fix it

PR #269, opened 2026-08-29
[develop] misc fixes

CajunAvenger:
putting one off fixes here so i don't forget about them


CajunAvenger:
* fixed maindfc not transferring notes
* fixed a linebreak issue in class special_text
* added borderless support to Adventure
* updated Name Check, regrouped as MtG Name Check
* added MSEM and Revolution Name Checks
* updated hint on all three of those to match the other exporters
* added transformation and partial dfc module support to Godzilla, and un-deprecated it. position hint and installer put to 1 before Godzilla Walkers. i'll probably come back and add transform notches and maybe bigger base node in 4.1
* fixed a misplaced dev file in dfc module

CajunAvenger:
* remove card_full_name_default it seemed to be redundant
* change card_full_name to be a hook where a card can redefine its base full name, such as a Split card being A//B or the Checklist Card being Checklist (though the two_name shape frames are covered by default
* add card_full_name_script as the script for card.full_name. this calls card_full_name(), then checks if there's a relevant DFC face to also call card_full_name() on to append to the full name. rendered as "Front // Back" and "Back (Front)"

PR #272, opened 2026-08-29
[develop] ERR Pixel frame (real)

CajunAvenger:
some different exe version snuck into the last one. i don't trust its not gonna be a pain and this is easier than figuring out how to command line the file out of there

template'd version of ERR's pixel frame
comes loaded with the Revolution pixel rarity symbol options and a little generator script to create your own
would it have been easier to just let people add the two images it uses to blend those instead
probably


G-e-n-e-v-e-n-s-i-S:
font files weren't added to Magic - Fonts folder

PR #284, opened 2026-09-02
add option to hide back faces from card list

G-e-n-e-v-e-n-s-i-S:
this one warrants some testing


CajunAvenger:
this will warrant a 2.6.1 tbh. would have been fair on the previous ones too

CajunAvenger:
i think a sort "X Hidden (Y back faces)" works fine for the tooltip

G-e-n-e-v-e-n-s-i-S:
alright, I'll add the pack tab scroll bar, and the upgrade to 16 potential card links in here as well, to make it a proper 2.6.1

G-e-n-e-v-e-n-s-i-S:
I set the max height of the "Pack selection" box to about 12 rows on the main tab, and 18 rows in the custom pack dialog. tell me if you want different numbers. Also, I'm not too sure how the "Totals" box behaves. can it also grow super large and need a scroll bar?

CajunAvenger:
Pack sizes look good~

Yes the totals box will also grow. There are two "tiers" of Pack elements; a "pack item" is the very baseline definitions that define a set filter, eg Rare or Mythic, and these are shown in the Custom Pack builder and the Totals box; and a "pack type" which puts multiple pack items together, both in the sense of "14% mythic, 86% rare" and "10 common, 3 uncommon, 1 rare", and those appear in the Pack selection window. Things like "Guaranteed Legendary" would end up on the Totals list unless an option is set to prevent it to

CajunAvenger:
noting that Meld Results are not hidden with this. perfectly fine with telling people tough luck if you don't want to deal with adding that in there

CajunAvenger:
(if it's even possible) do we want to change how Copy/Cut shortcuts work with back faces hidden? It does feel stranger to copy only the front face when you're displaying them as one card.

playing around with it on old set files and not seeing any problems. Selecting and editing back faces just fine, the back face isn't being eaten if you unlink or cut its front face from the file. Doing a bunch of linking and unlinking and haven't seen the card list break from it

G-e-n-e-v-e-n-s-i-S:
> noting that Meld Results are not hidden with this. perfectly fine with telling people tough luck if you don't want to deal with adding that in there

yeah that's gonna be considerably more involved, since it's not game agnostic, I'd have to implement a way for the game to specify what cards constitute a back face

G-e-n-e-v-e-n-s-i-S:
> (if it's even possible) do we want to change how Copy/Cut shortcuts work with back faces hidden? It does feel stranger to copy only the front face when you're displaying them as one card.

I can add a separate option to always copy the other face of a card

CajunAvenger:
image metadata doesn't need to be holding 16 (or even more) blank linked_card/relation fields

i'll check msem/rev for the most used token, i don't know what you'd use more than 16 links for other than like linking Clue to every single Investigate card

CajunAvenger:
Link Window still says Choose up to 4 to link, linking more works fine
buttons are working again

token trawl:
highest token of any kind: 79 Foretell
highest token that's not a rules reminder: 58 Soldier Army
highest token that's not crazy pants: 29 Clue
rules reminders we probably don't need to worry about
looks like we're still on hard coded fields, and i know the people with these 30+ tokens would love to be able to assign unique token arts more reliably, so i'd say 32 is generous and 64 isn't needed at all

CajunAvenger:
quirks when forcing weird links:
Meld Base / Anything adds the Meld bar to Anything, but only shares the card number if it's a Meld Result or Back Face
Any combination of Front Face and Meld Base makes a mess of card numbers and so has been forcibly blocked
Any combination of Back Face and Meld Result will both be treated as Front Faces, and not share numbers. We could similarly block that if we want to maintain that there's always a front/back pair or there isn't, back/back just doesn't cause problems the way front/front does.

G-e-n-e-v-e-n-s-i-S:
I did a minimal amount of testing, you can merge this whenever

PR #288, opened 2026-09-02
Pack cleanup

CajunAvenger:
current progress on the pack tab cleanup

There is one hopefully-easy change to the Packs code that would be nice; You are able to daisy chain filter properties, but only the last one is actually applied. I suspect [this line](https://github.com/G-e-n-e-v-e-n-s-i-S/MagicSetEditor2/blob/6c239c5531556bea6fd1561fba78592b20b2f3c1/src/data/pack.cpp#L112) is causing the filter to only run on set.cards, rather than the previous iteration's array of filtered cards. i've been tinkering with the source code but no luck yet.


CajunAvenger:
aight looks like the thing i wanted is gonna be a lot more involved so i'll just do it the long way instead

CajunAvenger:
1. redo the base rarity packs to support Showcases: slots are weighted so that individual cards show up at the same rate, and cards with Showcases linked via Standard/Showcase use their Showcase printing 20% of the time
2. add Standard/Showcase to the default links
3. add set fields to set two specialized notes that are checked on the slots Note 1 and Note 2, to build your own bonus sheet.
4. add noted slotting, nbslotting, and DFC slotting set fields, which determine if those notes, !nbslot note, and DFCs are allowed in the other slots or only the specialized ones
5. add partition boundary set field, to define what partitions show up in the standard pack slots
6. add some missing collation options like DFC, legendary slots, planeswalker slots
7. added is_standard_pack_card() as a more thorough check to decide if a card should appear in boosters
  - skip masterpieces
  - skip back faces
  - skip out of boundary partitions
  - skip dfcs if uniquely slotted (whitelist variable for DFC slots)
  - skip tokens
  - skip basic land
  - skip special rarity
  - skip promos
  - skip showcases
  - skip noted cards if uniquely slotted

i still need to review shifted slots, i'm sure there's someone whos been using that for 20 years who will rise from the grave to complain if i remove them entirely
also the Totals panel has been completely weirded, i might need to review the source code for what's happening there cause that doesn't make any sense rn

G-e-n-e-v-e-n-s-i-S:
> You are able to daisy chain filter properties

what do you mean by this? each PackType can only have one filter. In the code that you linked, I don't see it looping over filters, there is only one.

CajunAvenger:
> what do you mean by this? each PackType can only have one filter. In the code that you linked, I don't see it looping over filters, there is only one.

Each PackType can only have one filter, but resolving one Pack Card can drill through several PackTypes. I was hoping to have the filter apply at each step so you could just define `with showcase, without showcase, is showcase` and not have to do it for each rarity. Part cause it's unwieldy already and part because it starts to spiral out of control if you want to mix in DFCs or Shifted cards into that.

However, the way it works is each PackType just generates the cards that fit that filter at the very start and holds onto them, then gets a plain number of cards to pull from that array at the very end. I think I could have written something to pass filters along, but it would have been ugly and taken awhile, so I'll just go with the uglier pack options.

PR #290, opened 2026-09-06
add stroke radius/color/blur to fonts module

CajunAvenger:
working my way down the template list


CajunAvenger:
goodness i did not realize there were so many things using this, i wish i'd done this before launch

PR #298, opened 2026-09-16
MSE 2.6.4

G-e-n-e-v-e-n-s-i-S:
- add retry logic to IO operations. normally this shouldn't cause any slowdowns on normal drives, but it warrants being tested, it's a big change. it should hopefully clear some of the problems with cloud synced drives.
- run update_cards_scripts after installing updates (otherwise, cards are then saved with the new stylesheet version without ever having the scripts applied to them.)
- take font size into account when drawing checklist styling fields
- make `direction` text style property scriptable


CajunAvenger:
getting similar update times on 2.6.3 and 2.6.4 👍

PR #299, opened 2026-09-21
[4.1.0] level module redux

CajunAvenger:
* leveler module go
* add up to 6 combined_editor levels to whatever
* arrows, ptboxes, station, loyalty; extendable so templates can add their own new shapes without a lot of fuss
* handles the non-leveler pt field, but not loyalty/defense, templates that want those can add them style-side
* levels are colored by color_multiply, and allow for alternate level shapes and canvas areas similar to loyalty module
* originally coded the PT fields as minor offsets from the PT box but that's so ugly for the showcase frames that use canvas-size PT fields, so added a compatibility script for those
* this is like 1000 lines it might not make sense to anyone else and need some cleaning up done
* adding at least to mainframe and battles before launch, its only on a test frame and Fables right now for proof of concept


CajunAvenger:
Possible adjustments
* The HD scalar is determined by lesser dimension/375; but that can scrungle images a bit on the 744 frames, maybe it should be rounding that up to 2.
* Currently the module sets `flavor_bar_disabled_1 := { leveler_enabled() }` because the flavor bar is still active even if the second textbox isn't flavor text, might be a change to make to the flavor bar module instead
* Changing variables to functions:
  * coordinates arrays should def be arrays so they're concatable, otherwise adding new shapes or adjusting particular shapes is more annoying. They could return functions instead of numbers though, or i guess we are technically able to concat *functions* if we're careful about the inputs and don't mind making the user handle the return value
  * multi_array can go either way, but you can functionally script it by redefining it within the script calling color_multiply which I think reads nicer
* "missing" offset scripts
  * There's no width/height offset for containers/ptboxes; they can be several different dimensions so should be done by coords arrays instead
  * card.level_N and its label are enforced the same width to stay centered, i don't see value in differing that but maybe someone does
  * there's not offsets for individual levels but the global one is passed the level parameter you can add them that way
* Fonts: i had include filed: the font definitions to make expanding easier but maybe those should be each defined so you can modify them per level.

CajunAvenger:
and i guess some notes for "what i'm looking for"
* is this module sufficient for getting added to the showcase frames, including functionally taking over as the "PT module"
* what changes/dials/new features would you want to get it there or make it better
* if you don't want this mass-added to them that's fine, but i would like to get it to "can be added by request" at least
* stretch goals i'm looking at is adding additional card_field includes for separators between the levels and boxes under the levels, as alternatives to leveler_multiply. this could probably be used in place of insert_imaging stamp rims if you'd rather avoid that, where you just have a translucent overlay, that might end up looking better on the really dark textboxes of some showcases anyway.
* just realized i am gonna need to write a special_text script for this. probably just write it as special_text so most users don't have to worry about adding `special_text := {leveler_special_text()}`, i'm probably the other person who'd make a frame that needs to variant it

